PR #4, opened 2019-08-13
Apply admin: Create new exercise

alexbokii:
**Change log:**

1) Router - add a new route for a new exercise
2) Store - add a module for a new vacancy
3) Create 2 helpers - one to save a new vacancy to a Firestore, and one temporary to create a random ref number (it will be removed in the future but for now refs are required by a dashboard component)
4) Create components for main create new exercise page and for create the name of the exercise form.


tests/unit/components/NewExercise/NewExerciseName.spec.js
```diff
@@ -0,0 +1,48 @@
+import Vuex from 'vuex';
+import { shallowMount, createLocalVue } from '@vue/test-utils';
+import NewExerciseName from '@/components/NewExercise/NewExerciseName';
+
+const localVue = createLocalVue();
+localVue.use(Vuex);
+
+describe('CreateExercise view', () => {
```
ollietreend: Should be `describe('components/NewExercise/NewExerciseName', ...`

tests/unit/components/NewExercise/NewExerciseName.spec.js
```diff
@@ -0,0 +1,48 @@
...
+
+    expect(saveName).toHaveBeenCalled();
+  });
+
+  it('dispatches action to store', () => {
+    wrapper.setData({ name: 'testName' });
+    wrapper.vm.saveName();
+    expect(actions.setExerciseTitle.mock.calls.length).toBe(1);
+    expect(actions.setExerciseTitle.mock.calls[0][1]).toBe('testName');
+    expect(wrapper.emitted().submitted).toBeTruthy();
```
ollietreend: I think 'it emits `submitted` event' is probably another test. Or this test name should include it: 'it dispatches action to store and emits submitted event'.

It's not clear that we're testing for the emitted event right now.

tests/unit/helpers/saveNewExercise.spec.js
```diff
@@ -0,0 +1,43 @@
...
+import { firestore } from '@/firebase';
+
+jest.mock('@/firebase', () => {
+  const firebasemock = require('firebase-mock');
+  const mockfirestore = new firebasemock.MockFirestore();
+  const mocksdk = new firebasemock.MockFirebaseSdk(
+    // use null if your code does not use RTDB
+    () => null,
+    // use null if your code does not use AUTHENTICATION
+    () => null,
```
ollietreend: This mock gets simpler if we simply use the Firestore object which is already initialised in `@/firebase`:

```js
import { firestore } from '@/firebase';
```

(relates to my other comment on `src/helpers/saveNewExercise.js`)

tests/unit/journeys/signin.spec.js
```diff
@@ -63,5 +63,12 @@ describe('Sign in journey', () => {
         expect(subject.vm.$route.path).toBe('/dashboard');
       });
     });
+
+    describe('when going to the create new exercise page', () => {
+      it('can access the create new exercise', () => {
```
ollietreend: This doesn't read very well:

> "when going to the create new exercise page, it can access the create new exercise"

It could be more readable as:

> "when going to the create new exercise page, **it can access the page**"

tests/unit/store/createExercise.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    beforeEach(() => {
+      context = {
+        commit: jest.fn(),
+        getters,
+        state,
+      };
+    });
+
+    describe('setExerciseTitle', () => {
+      it('commits setTitle', () => {
```
ollietreend: Related to my other comment in `store/createExercise.js`:

If the `if()` statement is important in the action, we should test for it here.

i.e. there are two routes through this action, so we should test both scenarios

tests/unit/views/CreateExercise.spec.js
```diff
@@ -0,0 +1,51 @@
+import Vuex from 'vuex';
+import { shallowMount, createLocalVue } from '@vue/test-utils';
+import CreateExercise from '@/views/CreateExercise';
+import NewExerciseName from '@/components/NewExercise/NewExerciseName';
+import saveNewExercise from '@/helpers/saveNewExercise';
+import VueRouter from 'vue-router';
+
+jest.mock('@/helpers/saveNewExercise', () => jest.fn());
```
ollietreend: The real implementation of `saveNewExercise` returns a Promise, so the mock should return the same.

Something like this should probably work fine:

```js
jest.fn().mockResolvedValue()
```

because I don't think the return value of the Promise is used for anything (?)

src/router.js
```diff
@@ -25,6 +26,15 @@ const router = new Router({
         title: 'Dashboard',
       },
     },
+    {
+      path: '/exercise/new',
+      name: 'createExercise',
+      component: CreateExercise,
```
ollietreend: I think there's some mixed terminology in this PR – it's visible here, but it exists throughout. In some places we use the term 'New Exercise' whereas elsewhere we've used 'Create Exercise'.

I'm not sure if anything specifically needs to change, but it's just something for us to think about if it hasn't already been thought about.

src/components/NewExercise/NewExerciseName.vue
```diff
@@ -0,0 +1,53 @@
+<template>
```
ollietreend: This might be a step too soon – but when we discuss how our components become reusable, I can imagine us wanting to move this to `components/Exercise/Name.vue` since I think this view can be used for both our Create and Edit / Update views.

tests/unit/views/CreateExercise.spec.js
```diff
@@ -14,10 +14,17 @@ localVue.use(VueRouter);
 let store, wrapper, router;
 
 beforeEach(() => {
+  const getters = {
+    exerciseData: jest.fn().mockResolvedValue({ title: 'Test' }),
```
ollietreend: Getters should never return Promises – so in this case, I think you should be [mocking a return value](https://jestjs.io/docs/en/mock-function-api#mockfnmockreturnvaluevalue) rather than a resolved Promise value.

So this mock should be defined as:

```js
jest.fn().mockReturnValue({ title: 'Test' })
```

instead of:

```js
jest.fn().mockResolvedValue({ title: 'Test' })
```

PR #5, opened 2019-08-14
Add type component to the application form

alexbokii:
**Change log**

1. Add new component for ExerciseType
2. Include it into CreateExercise view and make it visible if title is available in a store
3. When type is added, reset state

### DIFF
--- a/src/components/NewExercise/NewExerciseType.vue
+++ b/src/components/NewExercise/NewExerciseType.vue
@@ -29,7 +29,7 @@
               id="non-legal"
               v-model="type"
               class="govuk-radios__input"
-              name="legal"
+              name="non-legal"
               type="radio"
               value="non-legal"
             >
--- a/src/components/NewExercise/NewExerciseType.vue
+++ b/src/components/NewExercise/NewExerciseType.vue
@@ -99,12 +99,10 @@ export default {
       type: '',
     };
   },
-  computed: {
+  methods: {
     ...mapActions('createExercise', [
       'setExerciseType',
     ]),
-  },
-  methods: {
     saveType () {
       this.$store.dispatch('createExercise/setExerciseType', this.type);
       this.$emit('submitted');
--- a/src/components/NewExercise/NewExerciseType.vue
+++ b/src/components/NewExercise/NewExerciseType.vue
@@ -1,0 +1,114 @@
+<template>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <fieldset class="govuk-fieldset govuk-!-padding-bottom-9">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h2 class="govuk-fieldset__heading">
+            Type of exercise
+          </h2>
+        </legend>
+        <div class="govuk-radios govuk-radios">
+          <div class="govuk-radios__item">
+            <input
+              id="legal"
+              v-model="type"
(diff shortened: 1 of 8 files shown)


src/components/NewExercise/NewExerciseType.vue
```diff
@@ -0,0 +1,114 @@
...
+import { mapActions } from 'vuex';
+
+export default {
+  data: function () {
+    return {
+      type: '',
+    };
+  },
+  computed: {
+    ...mapActions('createExercise', [
```
ollietreend: Vuex actions [should be mapped to the component as methods](https://vuex.vuejs.org/guide/actions.html#dispatching-actions-in-components) rather than computed properties. So `mapActions` should be used inside the `methods` definition in our component:

```
export default {
  methods: {
    saveType() { /* ... */ },
    ...mapActions( /* ... /* ),
  }
}
```

This is because Vuex actions are the functional equivalent of Vue component methods. Whereas Vuex getters are the equivalent of Vue component computed properties, so do map to `computed`.

src/views/CreateExercise.vue
```diff
@@ -1,29 +1,41 @@
...
 export default {
   components: {
     NewExerciseName,
+    NewExerciseType,
+  },
+  computed: {
+    ...mapGetters('createExercise', [
+      'exerciseData', 'exerciseTitle',
+    ]),
+    ...mapActions('createExercise', [
```
ollietreend: The same applies here – `mapActions` should be in `methods` not `computed`.

tests/unit/store/createExercise.spec.js
```diff
@@ -24,6 +25,26 @@ describe('store/createExercise', () => {
...
+    describe('setType', () => {
+      it('sets setType in the state', () => {
+        const type = 'Test Type';
+        mutations.setType(state, type);
+
+        expect(state.newExercise.type).toBe(type);
+      });
+    });
+
+     describe('resetExercise', () => {
```
ollietreend: The indentation of this `describe` block is off – it looks like it's spaced in by 5 chars instead of 4.

src/store/createExercise.js
```diff
@@ -5,23 +5,43 @@ const module = {
...
+        reference: generateRandomReference(1, 100),
+      };
+    },
   },
   actions: {
     setExerciseTitle({ commit }, name) {
       commit('setTitle', name);
     },
+    setExerciseType({ commit }, type) {
+      commit('setType', type);
```
ollietreend: Since these actions are just as wrappers to perform commits (mutations), we might be able to simplify this by using the mutations directly in our component. This would cut down on boilerplate code (and associated tests) because we wouldn't need these actions to exist.

Just as we have used `mapActions`, Vuex allows us to use `mapMutations` to [perform mutations from our components](https://vuex.vuejs.org/guide/mutations.html#committing-mutations-in-components).

src/views/CreateExercise.vue
```diff
@@ -1,29 +1,41 @@
 <template>
   <div>
     <NewExerciseName
+      v-if="!exerciseTitle"
       ref="newExerciseNameComponent"
+    />
+    <NewExerciseType
+      v-else
```
ollietreend: This works for now, but I think it could get a bit unwieldy once we have more questions to ask on the form.

Let's talk about this because I think the next step is to come up with the concept of 'steps' – and I think we already had something in the old application form which could be useful here.

PR #6, opened 2019-08-15
Add text field component for num of vacancies

alexbokii:
**Change log:**
1) Add component TextField
2) Create a new view for an "About the role" form
3) Use new TextField component to display Immediate start and Future start fields.

! to get to this component, go to http://localhost:8080/exercise/new/information-about-role.
As this journey should start on a View Vanancy page and View Vanancy page doesn't exist yet, we display form in a separate view which will be connected to View Vanancy page in the future.

### DIFF
--- a/src/components/Form/TextField.vue
+++ b/src/components/Form/TextField.vue
@@ -2,36 +2,51 @@
   <div class="govuk-form-group">
     <label
       :for="id"
-      class="govuk-heading-m" >
-      {{label}}
+      class="govuk-heading-m"
+    >
+      {{ label }}
     </label>
     <span 
       v-if="hint"
       id="event-name-hint" 
-      class="govuk-hint">
-      {{hint}}
+      class="govuk-hint"
+    >
+      {{ hint }}
     </span>
     <input 
-      :class="[ narrow ? 'govuk-input--width-5 govuk-input' : 'govuk-input' ]"
-      id="key" 
+      id="key"
+      v-model="textFieldValue" 
+      :class="[ (type === 'narrow') ? 'govuk-input--width-5 govuk-input' : 'govuk-input' ]" 
       name="key" 
-      type="text" 
-      v-model="textFieldValue">
+      type="text"
+    >
   </div>
 </template>
 
 <script>
-  export default {
-    data() {
-      return {
-        textFieldValue: ''
-      };
+export default {
+  props: {
+    label: {
+      default: '',
+      type: String,
     },
-    props: {
-      label: String,
-      hint: String,
-      id: String,
-      narrow: Boolean
-    }
-  }
+    hint: {
+      default: '',
+      type: String,
+    },
+    id: {
+      default: '',
+      type: String,
+    },
+    type: {
+      default: '',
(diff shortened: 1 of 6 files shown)


src/components/Form/TextField.vue
```diff
@@ -0,0 +1,52 @@
...
+    </label>
+    <span 
+      v-if="hint"
+      id="event-name-hint" 
+      class="govuk-hint"
+    >
+      {{ hint }}
+    </span>
+    <input 
+      id="key"
```
ollietreend: Should this be `:id="id"` so it dynamically matches the `:for="id"` on the label?

src/components/Form/TextField.vue
```diff
@@ -0,0 +1,52 @@
...
+  <div class="govuk-form-group">
+    <label
+      :for="id"
+      class="govuk-heading-m"
+    >
+      {{ label }}
+    </label>
+    <span 
+      v-if="hint"
+      id="event-name-hint"
```
ollietreend: If this `id` attribute is hardcoded, it'll cause duplicate IDs on the same page.

Do we need an ID on this element? Or can we make it dynamic by using the `id` prop and appending 'hint'? e.g. `:id="id + '_hint'"`

tests/unit/components/Form/TextField.spec.js
```diff
@@ -0,0 +1,59 @@
...
+         expect(wrapper.find('.govuk-hint').text()).toBe('Hint for the label!');
+      }); 
+
+      it('does not show hint if hint is not passed', () => {
+        expect(wrapper.find('.govuk-hint').exists()).toBe(false);
+      }); 
+    });
+
+    describe('id', () => {
+      it('sets the label for attribute', () => {
```
ollietreend: Could there also be a test that the `<input>` has the correct `id` attribute?

tests/unit/components/Form/TextField.spec.js
```diff
@@ -0,0 +1,59 @@
...
+        wrapper.setProps({ id: 'my_unique_key' });
+        expect(wrapper.find('label').attributes().for).toBe('my_unique_key');
+      }); 
+    });
+
+    describe('type', () => {
+      it('sets CSS class is narrow is true', () => {
```
ollietreend: Typo:

> sets CSS class **if** narrow is true

i.e. change 'is' to 'if'

tests/unit/components/Form/TextField.spec.js
```diff
@@ -0,0 +1,59 @@
...
+        wrapper.setProps({ type: 'narrow' });
+        expect(wrapper.find('input').attributes().class).toContain('govuk-input--width-5');
+      });
+
+      it("doesn't set CSS class is narrow is not passed", () => {
```
ollietreend: Same typo here – change 'is' to 'if'

src/components/Form/TextField.vue
alexbokii: I've just realised, we don't need id at all for this <span>, so I'll remove it for now

PR #7, opened 2019-08-16
Add view vacancy page

alexbokii:
**Change log:**

1. Update router to contain child routes for New Exercise page
2. Add ViewExercise view
3. Change the order of components visible to the user as per prototype


src/router.js
```diff
@@ -5,8 +5,12 @@ import store from '@/store';
 // Views
 import Dashboard from '@/views/Dashboard';
 import SignIn from '@/views/SignIn';
+import ViewExercise from '@/views/ViewExercise';
+// form
 import CreateExercise from '@/views/CreateExercise';
-import InformationAboutRole from '@/components/NewExercise/InformationAboutRole';
+import NameOfExercise from '@/components/NewExercise/NameOfExercise';
```
tatyree: Don't have to do it for this PR, but can we start to lazy load the route components, please?  
https://router.vuejs.org/guide/advanced/lazy-loading.html#grouping-components-in-the-same-chunk

You can do it directly in the route stanza like this: 

```
        {
          path: 'name-of-exercise',
          component: () => import('@/components/NewExercise/NameOfExercise'),
          meta: {
            title: 'Name Of Exercise',
          },
        },
```

alexbokii: Yes, great idea! It's a quick change, so I can do this now :)

tests/unit/views/ViewExercise.spec.js
```diff
@@ -0,0 +1,27 @@
+import ViewExercise from '@/views/ViewExercise';
+import { shallowMount, createLocalVue } from '@vue/test-utils';
+import Router from 'vue-router';
+
+const localVue = createLocalVue();
+localVue.use(Router);
+
+describe('ViewExercise view', () => {
```
tatyree: Don't need the second 'view' here.  'ViewExercise' is sufficient.  Won't block the PR over it, though.

PR #8, opened 2019-08-19
Scaffold the Eligibility view

LaurenQurashi:
TODO:
  - Plug the form into the database.
  - Add Ollie's reusable components when they're ready.
  - Have the submit button move to the next step.

This was copied and pasted directly from the prototype, and the happy path tests were adapted from the ViewExercise template.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -44,6 +44,7 @@ const router = new Router({
           component: Eligibility,
           name: 'eligibility',
           meta: {
+            requiresAuth: true,
             title: 'Eligibility',
           },
         },
--- a/src/router.js
+++ b/src/router.js
@@ -4,6 +4,7 @@ import store from '@/store';
 
 // Views
 import Dashboard from '@/views/Dashboard';
+import Eligibility from '@/views/Eligibility';
 import SignIn from '@/views/SignIn';
 import ViewExercise from '@/views/ViewExercise';
 // form
@@ -40,7 +41,7 @@ const router = new Router({
       children: [
         {
           path: 'eligibility',
-          component: () => import('@/views/Eligibility'),
+          component: Eligibility,
           name: 'eligibility',
           meta: {
             title: 'Eligibility',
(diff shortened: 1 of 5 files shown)


src/router.js
```diff
@@ -38,6 +38,14 @@ const router = new Router({
         title: 'Create new exercise',
       },
       children: [
+        {
+          path: 'eligibility',
+          component: () => import('@/views/Eligibility'),
```
alexbokii: As we can't test this lazy loading at the moment, can we please import components as for all other routes and add tests?

ollietreend: Dynamic route imports aren't compatible with the test runner we use (Jest), meaning we can't test the route configuration.

This is a known limitation noted [in the Vue docs](https://vue-test-utils.vuejs.org/guides/#testing-single-file-components-with-jest):

> **Note:** `vue-jest` currently does not support all the features of `vue-loader`, for example custom block support and style loading. In addition, some webpack-specific features such as code-splitting are not supported either. To use these unsupported features, you need to use Mocha instead of Jest to run your tests, and webpack to compile your components. To get started, read the guide on [testing SFCs with Mocha + webpack](https://vue-test-utils.vuejs.org/guides/#testing-single-file-components-with-mocha-webpack).

So for now, we'll need to keep this as a static import (i.e. the same as the other routes).

src/views/Eligibility.vue
```diff
@@ -0,0 +1,527 @@
...
+                <label
+                  class="govuk-label govuk-checkboxes__label"
+                  for="citizenship-conditional"
+                >
+                  Advocate (enrolled with the Scottish bar)
+                </label>
+              </div>
+              <div class="govuk-checkboxes__item">
+                <input
+                  id="citizenship-conditional"
```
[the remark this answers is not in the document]
LaurenQurashi: These are going to be replaced by your reusable components shortly, so as per our discussion, we'll leave them for now.

src/router.js
LaurenQurashi: done!

LaurenQurashi: done

PR #9, opened 2019-08-20
Scaffold the About The Role view

LaurenQurashi:
TODO:
  - Plug the form into the database.
  - Add Ollie's reusable components when they're ready.

This was copied and pasted directly from the prototype, and the happy 
path tests were adapted from the ViewExercise template. Tests include 
happy path, and unauthenticated user redirect tests.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -3,6 +3,7 @@ import Router from 'vue-router';
 import store from '@/store';
 
 // Views
+import AboutTheRole from '@/views/AboutTheRole';
 import Dashboard from '@/views/Dashboard';
 import Eligibility from '@/views/Eligibility';
 import SignIn from '@/views/SignIn';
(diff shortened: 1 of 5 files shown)


alexbokii:
Looks good to me @LaurenQurashi! 
The only thing, can we add a test for this page, which checks that component is rendered and the page has h1?

ollietreend:
Looks good @LaurenQurashi, aside from @alexbokii's comment 😄

We're just missing a test for the view component itself (`views/AboutTheRole.vue`). I think that was included in your [PR for Eligibility](https://github.com/jac-uk/apply-admin/pull/8), so you should be able to use that as a starting point.

tatyree:
@ollietreend I checked @LaurenQurashi work re. adding the test and went ahead and approved to unblock her.

PR #12, opened 2019-08-23
Add Date and Time components

alexbokii:
**Change log:**

1) Date Component - add component, helpers and tests
2) Time Component - add component, helpers and tests

These two components will be combined into one component - DateTime in the next PR


LaurenQurashi:
Looks great to me! I'll let Ollie approve as I'm not too confident about what I'm doing 😂

tests/unit/helpers/Form/zeroPad.spec.js
```diff
@@ -0,0 +1,15 @@
+import zeroPad from '@/helpers/Form/zeroPad';
+
+describe('helpers/zeroPad', () => {
```
ollietreend: As mentioned above – I'd really like for `zeroPad` to accept `0` as a valid input and return `'00'`.

tests/unit/helpers/Form/zeroPad.spec.js
```diff
@@ -0,0 +1,15 @@
...
+  it('adds zero in the beginning is the lengths of the number is one', () => {
+    expect(zeroPad(3)).toBe('03');
+  });
+
+  it('does not modify string with the length equal to two', () => {
+    expect(zeroPad(33)).toBe('33');
+  }); 
+
+  it('returns null if value is undefined', () => {
```
ollietreend: This test description is slightly misleading because we're passing in an empty string, not `undefined`.

Could we rename it to 'returns null if value is an empty string'?

Additionally, we may want to test what happens when we pass in `null` values, since we know that's a possible input value for our use case – e.g. in `<DateInput />` if `this.day` is null, then our `dayInput` getter will pass `null` into this `zeroPad` helper.

src/components/Form/DateInput.vue
```diff
@@ -0,0 +1,195 @@
...
+
+<script>
+import validateDay from '@/helpers/Form/validateDay';
+import validateMonth from '@/helpers/Form/validateMonth';
+import validateYear from '@/helpers/Form/validateYear';
+import zeroPad from '@/helpers/Form/zeroPad';
+
+export default {
+  props: {
+    label: {
```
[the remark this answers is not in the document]
ollietreend: We've agreed to make the `label` prop optional across all components. This is consistent with the `hint` prop, which is also optional everywhere.

Marking this conversation as resolved.

src/components/Form/TimeInput.vue
```diff
@@ -0,0 +1,171 @@
...
+      },
+      set(value) {
+        this.minute = validateMinute(value);
+      },
+    },
+    dateConstructor() {
+      const hour = this.hour; 
+      const minute = this.minute;
+
+      if(hour === null  || minute === null) {
```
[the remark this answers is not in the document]
ollietreend: I've [added this to our Trello board](https://trello.com/c/wNyOgVcE/377-lint-rule-for-consistent-spacing-around-keywords-if-else-case-etc) in the 'Could have' column.

Marking this conversation as resolved.

src/helpers/Form/zeroPad.js
```diff
@@ -1,8 +1,13 @@
 const zeroPad = (val) => {
-  if(!val) {
+  if(val === 0) {
```
ollietreend: Let's simplify the implementation of this – rather than checking if the `val` is truthy and a number.

I think we can simplify this to:

- If `val` is not a number, return `null`
- If not, return a zero padded string

We don't need a special case for `0` because it's a number, so will have the same treatment as all other numbers.

tests/unit/helpers/Form/zeroPad.spec.js
```diff
@@ -9,7 +9,14 @@ describe('helpers/zeroPad', () => {
...
-  it('returns null if value is undefined', () => {
-    expect(zeroPad('')).toBe(null);
+  it('returns null if the value passed is not a number', () => {
+    expect(zeroPad('string')).toBe(null);
+    expect(zeroPad(undefined)).toBe(null);
+    expect(zeroPad(true)).toBe(null);
+    expect(zeroPad(null)).toBe(null);
+  });
+
+  it('returns "00" if "0" string is passed', () => {
```
ollietreend: This test description should change – we don't pass a string "0", it's a number.

How about: `returns "00" if number 0 is passed`

ollietreend:
Looks great @alexbokii – let's merge this! 👍

PR #16, opened 2019-08-28
Text Field Review

LaurenQurashi:
Refactored the Unit Tests for the TextField component. 

- Removed the `type` property from each Prop. 
- Reworded the tests to be clearer. 
- Removed the accidental installation of npx.


src/components/Form/TextField.vue
```diff
@@ -6,18 +6,18 @@
...
+    <input
       :id="id"
-      v-model="textFieldValue" 
-      :class="[ (type === 'narrow') ? 'govuk-input--width-5 govuk-input' : 'govuk-input' ]" 
-      :name="id" 
-      type="text"
+      v-model="text"
+      :class="inputClass + ' govuk-input'"
+      :name="id"
+      type="tel"
```
alexbokii: We won't always want to have type="tel", but as we agreed on changing this in next PR, it's not a problem.

alexbokii:
Looks good, well done @LaurenQurashi ! 😄

PR #20, opened 2019-08-29
Scaffold Create An Exercise

LaurenQurashi:
- Scaffolded the create an exercise page using the newly created components.
- Conducted a file clean up of some pages that were no longer needed, along with their corresponding tests. 

- We need to address the styling inaccuracies within the components. They do not match the prototype at the moment.


src/views/CreateAnExercise.vue
```diff
@@ -1,166 +1,81 @@
...
+          />
+        </RadioGroup>
+        <RadioGroup
+          id="add-information"
+          v-model="isMoreInfoNeeded"
+          label="Do you want to add more information about this exercise now?"
+          hint="You can add exercise contacts, shortlisting methods, timeline dates, or information from HMCTS. You can also do this later"
+        >
+          <RadioItem
+            value="yes"
```
ollietreend: Let's use boolean values for the 'yes' and 'no' responses here.

Since the value isn't a `String` but will be `Boolean`, you'll need to change your `value` attribute to be a JavaScript expression binding `:value`, like so:

```vue
<RadioItem
  :value="true"
  label="Yes"
/>
```

Without the `:` binding, it will simply be the string `"true"` which isn't as helpful to us.

PR #21, opened 2019-08-29
GOV.UK Checkboxes input component

ollietreend:
This PR adds the components `CheckboxGroup` and `CheckboxItem`. These behave almost exactly the same as the recently built `RadioGroup` and `RadioItem` components.


ollietreend:
@alexbokii I will add documentation for Radios and Checkboxes to `README.md` in a separate PR.

src/components/Form/CheckboxItem.vue
```diff
@@ -0,0 +1,90 @@
...
+    }
+    index++;
+  },
+};
+</script>
+
+<style scoped lang="scss">
+  // We need to wrap both the input & conditional content in one top-level element
+  // This is because Vue doesn't (yet) support components with multiple top-level elements (e.g. React Fragments)
+  // This results in our markup structure differing slightly from the officially documented markup for this component
```
LaurenQurashi: Not to be picky, as I'm only looking for things to comment about, but do we need these comments? 🤷‍♀

ollietreend: Good question @LaurenQurashi. I've included these comments here so that other developers working on the codebase can understand why custom styles are used on this component. Ideally custom styles wouldn't be needed here because styling is provided by the GOV.UK Frontend library.

If these styles were to be changed or removed, the visual impact would be subtle and could be easy to miss – but it would ultimately result in a broken visual design. And since our unit tests don't cover styling, they'd continue to pass.

That's why I felt it was worth adding those comments – so that I can explain why I 'broke the rules' of using custom styles to re-create GOV.UK styles, and why it's important for them to stay.

PR #25, opened 2019-08-30
Scaffold shortlisting methods

LaurenQurashi:
Added our components to the shortlisting methods page. Also added a pathway to our router. All tests are passing, and code has been linted. 

- Need to link it up to the AddExerciseContacts page via 'save and continue' button. 
- Need to have the 'save and continue' button point to the next page.

### DIFF
--- a/src/views/AddShortlistingMethods.vue
+++ b/src/views/AddShortlistingMethods.vue
@@ -1,0 +1,80 @@
+<template>
+  <main class="govuk-main-wrapper">
+    <div class="govuk-grid-row">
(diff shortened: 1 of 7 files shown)


src/views/AddShortlistingMethods.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <main class="govuk-main-wrapper">
```
alexbokii: This <main> element is not needed as we already have <main> in App.vue, which wraps all our pages. Can we please remove <main> from here?
(also I guess all other pages contain two main elements as well)

PR #27, opened 2019-09-02
Dummy Ref Number

LaurenQurashi:
- Added a dummy ref number in to the Create an Exercise page.

- It's a hard coded component. Copied the format Kellie used in the prototype but tried to tidy it up a bit so it's inline with our components. 

Please let me know if there's anything that needs changing as I wasn't too sure about how much I could clean up 🌻

### DIFF
--- a/src/views/CreateAnExercise.vue
+++ b/src/views/CreateAnExercise.vue
@@ -4,6 +4,23 @@
       <h1 class="govuk-heading-xl">
         Create an exercise
       </h1>
+      <div class="govuk-form-group">
+        <label class="govuk-heading-m">
+          Reference number
+        </label>
+        <span id="hint" class="govuk-hint">
+          This will be generated automatically when you save.
+        </span>
+        <input
+          class="govuk-input--width-5 govuk-input"
+          id="ref-number"
+          name="ref-number"
+          type="text"
+          placeholder="XXX"
+          disabled=""
+          style="background-color:#ccc"
+        />
+    </div>
       <TextField
         id="exercise-name"
         v-model="exerciseName"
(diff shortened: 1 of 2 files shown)


src/views/CreateAnExercise.vue
```diff
@@ -4,6 +4,23 @@
       <h1 class="govuk-heading-xl">
         Create an exercise
       </h1>
+      <div class="govuk-form-group">
+        <label class="govuk-heading-m">
```
ollietreend: We want our `<label>` to have a `for` attribute which points to the `<input>`. So in this case, where `<input id="ref-number">` we will need `<label for="ref-number">`.

src/views/CreateAnExercise.vue
```diff
@@ -4,6 +4,23 @@
...
+          Reference number
+        </label>
+        <span id="hint" class="govuk-hint">
```
ollietreend: This hint `<span>` doesn't have a very unique ID right now. To make this consistent with our `TextField` component, let's simply remove the `id` attribute from this.

src/views/CreateAnExercise.vue
```diff
@@ -4,6 +4,23 @@
...
+          This will be generated automatically when you save.
+        </span>
+        <input
+          class="govuk-input--width-5 govuk-input"
+          id="ref-number"
+          name="ref-number"
+          type="text"
+          placeholder="XXX"
+          disabled=""
+          style="background-color:#ccc"
```
ollietreend: The `name` attribute isn't needed on this field – let's remove it.

The disabled attribute should simply be `disabled` rather than `disabled=""` – we don't need to set it to an empty string.

Let's adapt the inline styles here (in the `style` attribute) into a more generic rule for disabled text inputs. We'll want to create a rule in the `main.scss` file.

PR #29, opened 2019-09-03
Added Button Routing and Tests.

LaurenQurashi:
- Added routing for the current three pages.
- Added tests to the Views folder for these pages. 
- Run lint, Tests pass.

### DIFF
--- a/src/views/AddExerciseContacts.vue
+++ b/src/views/AddExerciseContacts.vue
@@ -143,7 +143,9 @@
       </p>
 
       <button
+        ref="linkToAddShortlistingMethods"
         class="govuk-button"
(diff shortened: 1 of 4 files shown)


src/views/AddExerciseContacts.vue
```diff
@@ -143,7 +143,9 @@
       </p>
 
       <button
+        ref="linkToAddShortlistingMethods"
```
tatyree: Do you need the `ref`s for these?  No matter either way for the purposes of this PR, I just didn't see them when I looked at the source.

tatyree: Nevermind, just found it...

PR #33, opened 2019-09-05
Scaffold About the Vacancy

LaurenQurashi:
- Created an About the Vacancy page, in accordance with the V2.1 Prototype requirements. 
- Tests have been written and are passing.
- Linter has been run on the code.

-Need to review: 
   - The Appointments section --> The embedded salary box doesn't look like the prototype, and the input does not save anywhere.

- The Upload Documents section --> The input does not save anywhere.

### DIFF
--- a/src/views/AboutTheVacancy.vue
+++ b/src/views/AboutTheVacancy.vue
@@ -186,7 +186,6 @@
       <TextField
         id="immediate-start"
         v-model="immediateStart"
-        name="immediate-start"
         label="Immediate start"
         input-class="govuk-input--width-3"
         hint="These are also called Section 87 (S87) vacancies."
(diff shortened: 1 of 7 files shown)


src/views/AboutTheVacancy.vue
```diff
@@ -0,0 +1,347 @@
...
+        name="immediate-start"
+        label="Immediate start"
+        input-class="govuk-input--width-3"
+        hint="These are also called Section 87 (S87) vacancies."
+      />
+
+      <TextField
+        id="future-start"
+        v-model="futureStart"
+        name="future-start"
```
alexbokii: Do we need to pass name separately?
Looks like TextInput component uses id prop to create value for name attribute.

PR #38, opened 2019-09-09
Repeatable fields

alexbokii:
This PR adds the main component for Repeatablefields, and adds a reusable field for a SelectionDay to the Timeline page.
PR for repeatable Textinput is upcoming. 

**Change log:**

1) Add Repeatablefields component and tests
2) Create SelectionDay component
3) Integrate SelectionDay repeatable component into a Timeline page

### DIFF
--- a/src/components/RepeatableFields.vue
+++ b/src/components/RepeatableFields.vue
@@ -42,6 +42,7 @@ export default {
     },
     component: {
       required: true,
+      type: Object,
     },
     max: {
       required: false,
--- a/src/components/RepeatableFields.vue
+++ b/src/components/RepeatableFields.vue
@@ -4,16 +4,17 @@
       v-for="(row, index) in rows"
       :key="index"
       ref="repeatableField"
+      class="repeatableField"
     >
       <component
         :is="component"
         :row="row"
         :index="index"
       >
-        <template slot="removeButton">
+        <template v-slot:removeButton>
           <button
             type="button"
-            class="govuk-link govuk-link-delete"
+            class="govuk-link govuk-link-delete govuk-!-margin-bottom-4"
             @click.prevent="removeRow(index)"
           >
             Remove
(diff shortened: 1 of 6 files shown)


src/components/RepeatableFields.vue
```diff
@@ -0,0 +1,105 @@
...
+    >
+      <component
+        :is="component"
+        :row="row"
+        :index="index"
+      >
+        <template slot="removeButton">
+          <button
+            class="govuk-link govuk-link-delete"
+            @click.prevent="removeRow(index)"
```
ollietreend: Can we add attribute `type="button"` to this?

Without explicitly setting the type, `<button>`s default to `type="submit"`. This will cause problems when we implement save functionality, in which we'll wrap the entire page in a `<form>`.

I can demo this behaviour – but essentially what will happen is the 'Remove' button will be triggered when the user hits 'Enter' key on an input inside a repeatable field component. So hitting 'Enter' will delete the row. 😳

ollietreend: In fact – thinking about it. Should we simply be marking these up as `<a>` hyperlinks rather than `<button>`s?

They feel like buttons because they're performing an action on the page, not navigating to a different page. But I know the UX team is keen for these to be styled as links. Currently as `<button>` elements, they don't have a `cursor: pointer` on hover, which feels unusual.

src/components/RepeatableFields.vue
```diff
@@ -0,0 +1,105 @@
...
+          </button>
+        </template>
+      </component>
+    </div>
+    <div class="text-right">
+      <button
+        v-if="canAddRow"
+        class="govuk-link"
+        @click.prevent="addRow"
+      >
```
ollietreend: The same goes for this 'Add another' `<button>` as with the 'Remove' button.

ollietreend:
This looks really nice @alexbokii, and the tests look good. You've tested all the important functionality – add, remove, and the repeated component render.

PR #39, opened 2019-09-09
V2.1 Contacts

LaurenQurashi:
Edited the page to align with the Prototype V2.1. 

- Rearranged components and added any additional details to them. 
- Run linter on the code
- Run tests - they're all passing. 

Note: 
- I removed the name attribute from the TextField components on this page as it was mentioned earlier that they're not required. 
- I removed the href from the Add Another links as @alexbokii is working on them. The code that's there is just going to be a placeholder for when it's ready for proper functionality. 

- I can't seem to get the back button to work so if anyone knows what i'm doing wrong there, that would be great :)


src/views/Exercises/Edit/Contacts.vue
```diff
@@ -14,57 +21,41 @@
...
-      />
-
-      <TextField
-        id="equality-diversity-contact"
-        v-model="equalityDiversityContact"
-        name="equality-diversity-contact"
-        label="Equality and diversity contact"
+        id="exercise-mailbox"
+        v-model="exerciseMailbox"
+        label="Exercise mailbox"
```
ollietreend: If this field is for capturing an email address (I think it is..), then we should set `type="email"` on the `<TextField>` input.

This achieves a couple of things:

1. It'll enable built-in validation in the browser. If you type something which isn't an email address in the box, you'll get a validation prompt from the browser ([something like this](https://cms-assets.tutsplus.com/uploads/users/30/posts/25145/image/emaik.png)).
2. On touch-screen devices (phone, tablet, etc.), it'll usually customise the on-screen keyboard to make the `@` symbol more prominent / easier to find since the user is expected to enter an email address.

src/views/Exercises/Edit/Contacts.vue
```diff
@@ -89,24 +80,45 @@
...
+          value="welsh-government"
+          label="Welsh Government"
+        />
+        <CheckboxItem
+          value="other"
+          label="Other"
+        >
+          <TextField
+            id="other-text-input"
+            v-model="otherText"
```
ollietreend: This `v-model` field is probably better named as `otherAppropriateAuthority`, since this is what the 'other' field relates to.

PR #41, opened 2019-09-09
Edited Create Exercise page in line with Prototype V2.1

LaurenQurashi:
- Edited components to meet new requirements.
- Edited the save method to not record a type of exercise. 
- Edited the tests to not expect an exercise type. 

- Run linter on code. 
- All tests pass.

### DIFF
--- a/src/views/Exercises/New.vue
+++ b/src/views/Exercises/New.vue
@@ -13,10 +13,7 @@
           Create an exercise
         </h1>
 
-        <label
-          for="ref-number"
-          class="govuk-heading-m"
-        >
+        <label class="govuk-heading-m">
           Reference number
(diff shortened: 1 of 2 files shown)


src/views/Exercises/New.vue
```diff
@@ -2,66 +2,71 @@
...
-            id="ref-number"
-            class="govuk-input--width-5 govuk-input"
-            type="text"
-            placeholder="XXX"
-            disabled
-          >
-        </div>
+
+        <label
+          for="ref-number"
```
ollietreend: Let's drop the `for` attribute from this label, as it's no longer needed now that we don't have a `<input>` associated with it.

PR #42, opened 2019-09-10
Add repeatableTextFields

alexbokii:
This PR adds two repeatable text fields that can be used as an example for all future repeatable text fields.

**Change log**
1) Update style of the repeatable component so only the first label in repeatableContainer is visible
2) Create two repeatable text fields components for a contact page and integrate them into a page
3) Add a test that can be reused for all future repeatable text fields

### DIFF
--- a/src/components/RepeatableFields.vue
+++ b/src/components/RepeatableFields.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="repeatableContainer">
     <div
       v-for="(row, index) in rows"
       :key="index"
@@ -108,4 +108,15 @@ export default {
   .repeatableField .govuk-form-group {
     margin-bottom: 1em;
   }
+
+  .repeatableContainer .repeatableField:not(:first-child) label {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    padding: 0;
+    margin: -1px;
+    overflow: hidden;
+    clip: rect(0,0,0,0);
+    border: 0;
+  } 
 </style>
--- a/src/components/RepeatableFields.vue
+++ b/src/components/RepeatableFields.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="repeatableContainer">
+  <div>
     <div
       v-for="(row, index) in rows"
       :key="index"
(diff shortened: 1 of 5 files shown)


src/components/RepeatableFields.vue
```diff
@@ -108,4 +108,15 @@ export default {
   .repeatableField .govuk-form-group {
     margin-bottom: 1em;
   }
+
+  .repeatableContainer .repeatableField:not(:first-child) label {
```
ollietreend: I think we should be really cautious about making global CSS rules at this level. As it stands, this CSS rule has unwanted side effects:

- Every label inside the repeatable field component (aside from the first one) is hidden. This means date fields lose their 'Day', 'Month', 'Year' labels, for example:
  Add_Exercise_Timeline___Judicial_Appointments_Commission
- Since not all fields use `<label>` elements to display the label (e.g. date input uses `<legend>`), this leads to inconsistent behaviour between field types.

I feel quite strongly that the `RepeatableFields` component should only provide repeatable functionality, but it should not make any assumptions about children or interfere with their styling. Otherwise we risk adding CSS rules in several different places and it becoming hard to find which rules affect which elements.

How about we apply the hidden label styling as scoped CSS rules directly on the child components where it's required? i.e.

- `RepeatableFields/DraftingJudge.vue`
- `RepeatableFields/StatutoryConsultee.vue`

If we want to avoid repetition of the rules, perhaps we could use a specific class name in `main.scss` which is applied to those elements?

tests/unit/components/RepeatableFields/RepeatableTextFields.spec.js
```diff
@@ -0,0 +1,40 @@
...
+import DraftingJudge from '@/components/RepeatableFields/DraftingJudge';
+import StatutoryConsultee from '@/components/RepeatableFields/StatutoryConsultee';
+
+// add repeatable text component to this array
+const repeatableTextFields = [
+  ['DraftingJudge', DraftingJudge],
+  ['StatutoryConsultee', StatutoryConsultee],
+];
+
+describe('RepeatableTextFields', () => {
```
ollietreend: I think we could improve the readability of these tests if we use one `describe.each` for each component, rather than multiple `it.each` tests. Something like this:

```js
describe('Repeatable text fields', () => {
  describe.each(repeatableTextFields)('@/components/RepeatableFields/%s', (label, component) => {
    it('does something', () => {});
    it('does something else', () => {});
    describe('props', () => {
      // ...
    });
  });
});
```

I think this approach could be easier to understand when reading the test file, because the `.each` loop only happens once and very clearly shows that the same tests are applied to each component.

What do you think?

alexbokii: Done :)

src/components/RepeatableFields.vue
alexbokii: As we agreed with @ollietreend we won't hide any labels for now. So all styling changes are removed from this PR

PR #44, opened 2019-09-10
V2.1 timeline

LaurenQurashi:
Updated the Timeline page to match the Prototype V2.1

- Rearranged and edited components as required.
- Added navigation functionality to the buttons. 
- Added a location textfield component into the SelectionDay component.
- Code has been linted
- Tests are passing. 

Note: 

- What do we think about the v-Model naming? With the Situational Judgement and Critical Analysis tests sharing a field, it's been hard to come up with a comprehensive name for them. 

thanks 😎 💻 🖥 📱

### DIFF
--- a/src/views/Exercises/Edit/Timeline.vue
+++ b/src/views/Exercises/Edit/Timeline.vue
@@ -4,7 +4,6 @@
       <h1 class="govuk-heading-xl">
         Add exercise timeline
       </h1>
-
       <p class="govuk-body-l">
         You can return to this page later to add or change dates.
       </p>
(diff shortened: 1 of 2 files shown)


src/views/Exercises/Edit/Timeline.vue
```diff
@@ -36,33 +43,68 @@
       />
 
       <h2 class="govuk-heading-l">
-        Situational judgment test, Critical analysis test
+        Shortlisting dates
+      </h2>
+
+      <h2 class="govuk-heading-m">
```
ollietreend: In this case, our medium heading should probably be a `<h3>`. Visually it's smaller than the larger `<h2>` above, so hierarchically I think it should be a sub-heading to the main 'Shortlisting dates' heading.

PR #45, opened 2019-09-10
Create Edit page

alexbokii:
**Change log:** 

1) Add LoadingMessage component

### DIFF
--- a/src/views/Exercises/Edit.vue
+++ b/src/views/Exercises/Edit.vue
@@ -1,9 +1,40 @@
 <template>
   <div>
-    <RouterView />
+    <LoadingMessage
+      v-if="loaded === false"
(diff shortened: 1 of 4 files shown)


src/views/Exercises/Edit.vue
```diff
@@ -1,9 +1,39 @@
...
+    };
+  },
+  mounted() {
+    const id = this.$route.params.id;
+    
+    this.$store.dispatch('exerciseDocument/bind', id)
+      .then(() => {
+        this.loaded = true;
+      }).catch(() => {
+        this.loadFailed = true;
```
ollietreend: Please also throw the error which is caught here (it'll help when we have Sentry integration)

PR #53, opened 2019-09-12
Add View parent page

alexbokii:
This PR adds a parent Show page and empty templates for child components.

**Change log**
1) Create Navigation component 
2) Create Edit page that uses Navigation component and displays Overview child page by default

### DIFF
--- a/src/components/Page/Navigation.vue
+++ b/src/components/Page/Navigation.vue
@@ -2,7 +2,7 @@
   <nav :aria-label="label">
     <ul class="dwp-vertical-navigation">
       <li 
-        v-for="navItem in data"
+        v-for="navItem in items"
         :key="navItem.name"
       >
         <router-link 
@@ -20,7 +20,7 @@
 <script>
 export default {
   props: {
-    data: {
+    items: {
       required: true,
       type: Array,
     },
--- a/src/components/Page/Navigation.vue
+++ b/src/components/Page/Navigation.vue
@@ -1,0 +1,65 @@
+<template>
+  <nav :aria-label="label">
+    <ul class="dwp-vertical-navigation">
+      <li 
+        v-for="navItem in data"
+        :key="navItem.name"
+      >
+        <router-link 
+          class="nav-link"
+          :to="{name: navItem.name}"
+          :aria-current="isActive(navItem.name) ? 'page' : false"
+        >
+          {{ navItem.page }} 
+        </router-link>
+      </li>
+    </ul>
+  </nav>
+</template>
+
+<script>
+export default {
+  props: {
+    data: {
+      required: true,
+      type: Array,
+    },
+    label: {
+      required: false,
+      default: 'navigation',
+      type: String,
+    },
+  },
+  methods: {
+    isActive(name) {
+      return name === this.$route.name;
+    },
+  },
+};
+</script>
+
+<style scoped>
+  .dwp-vertical-navigation {
+    margin: 0 0 1.25em 0;
+  }
+
+  .dwp-vertical-navigation li {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    font-size: 1em;
+  }
+
+  .dwp-vertical-navigation a[aria-current="page"] {
+    color: #1d70b8;
+    font-weight: bold;
+    text-decoration: none;
+    border-left: 4px solid #1d70b8;
+    background-color: #f3f2f1;
+  }
+
+  .dwp-vertical-navigation li a {
+    padding: 0.625em;
+    display: block;
+  }
+</style>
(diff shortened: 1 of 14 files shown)


src/views/Exercises/Show.vue
```diff
@@ -0,0 +1,80 @@
...
+      navPages,
+    };
+  },
+  mounted() {
+    const id = this.$route.params.id;
+    
+    this.$store.dispatch('exerciseDocument/bind', id)
+      .then(() => {
+        this.loaded = true;
+        this.exercise = this.$store.state.exerciseDocument.record;
```
ollietreend: Rather than setting `this.exercise` on load, can this be changed to be a computed property which reads from the state? That way, we'll get live updates in the view whenever the record changes since the Vuex store is bound to Firestore.

tests/unit/views/Exercises/Show.spec.js
```diff
@@ -0,0 +1,67 @@
...
+  describe('template', () => {
+    describe('when loaded is false', () => {
+      it('renders LoadingMessage component', () => {
+        let wrapper = createTestSubject();
+        expect(wrapper.find({ ref: 'loadingMessageComponent' }).exists()).toBe(true);
+      });
+    });
+
+    describe('when loaded is true', () => {
+      it('renders LoadingMessage component', () => {
```
ollietreend: I think this description should be "does not render LoadingMessage component"?

tests/unit/views/Exercises/Show.spec.js
```diff
@@ -0,0 +1,67 @@
...
+        let wrapper = createTestSubject();
+        wrapper.setData({ loaded: true });
+        expect(wrapper.find({ ref: 'loadingMessageComponent' }).exists()).toBe(false);
+      });
```
ollietreend: Could we also test inside this `describe()` block that `RouterView` is rendered? (i.e. "it renders the child page")

This is probably also a good place to test that "it renders the Navigation component", because that also happens in this scenario. So maybe move from where it currently is.

tests/unit/components/Page/Navigation.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        expect(prop.default).toBe('navigation');
+      });
+
+      it('should be string', () => {
+        expect(prop.type).toBe(String);
+      });
+    });
+  });
+
+  describe('markup', () => {
```
ollietreend: For consistency with other tests, can we change this describe block from "markup" to "template"?

tests/unit/components/Page/Navigation.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    it('sets aria-label with label prop', () => {
+      let wrapper = createTestSubject({ data: navPages, label: 'MyTestLabel' });
+      expect(wrapper.find('nav').attributes('aria-label')).toBe('MyTestLabel');
+    });
+  });
+
+  describe('Accessibility', () => {
+    describe('aria-current attribute', () => {
```
ollietreend: Does this also belong inside the "markup"/"template" describe block? Rather than in its own separate block. It still describes behaviour of the template, the same as the `aria-label` test.

src/components/Page/Navigation.vue
```diff
@@ -0,0 +1,65 @@
...
+        </router-link>
+      </li>
+    </ul>
+  </nav>
+</template>
+
+<script>
+export default {
+  props: {
+    data: {
```
ollietreend: `data` is a bit of a vague name for the property. Could we name it something more descriptive?

I'm thinking maybe it could be called `items` – i.e. these are navigation menu items

tests/unit/views/Exercises/Show.spec.js
```diff
@@ -0,0 +1,85 @@
...
+        expect(wrapper.vm.exercise).toEqual(exercise);
+      });
+    });
+  });
+
+  describe('template', () => {
+    describe('when loaded is false', () => {
+      it('renders LoadingMessage component', () => {
+        let wrapper = createTestSubject();
+        expect(wrapper.find({ ref: 'loadingMessageComponent' }).exists()).toBe(true);
```
ollietreend: It might be cleaner to import the `LoadingMessage` component and use `wrapper.find(LoadingMessage)` here rather than using a `ref`. This would be consistent with the way it's done for `Navigation` and `RouterView`, and allows us to simplify the template markup by removing the `ref` attribute.

tests/unit/views/Exercises/Show.spec.js
```diff
@@ -0,0 +1,85 @@
...
+      });
+    });
```
ollietreend: We could benefit from having another test here: "when loaded is false > it does not render the RouterView"

That way we test the other side of our assumptions, and make it clear that the child page should never be rendered before loading has completed in the parent.

PR #56, opened 2019-09-13
Amended the Currency component.

LaurenQurashi:
We were receiving an error when using the currency component, due to it 
expecting a number rather than a string. 

- I've amended the component's Value Prop only, to expect a string. I've 
left the type as number in order to keep accessibility features that 
come with this. 

- I've amended the test to expect a string rather than a number to be 
more accurate (Not sure why the number ones were still passing tbh?)

Tests are passing and code has been linted

🌻

### DIFF
--- a/tests/unit/components/Form/Currency.spec.js
+++ b/tests/unit/components/Form/Currency.spec.js
@@ -62,8 +62,8 @@ describe('components/Form/Currency', () => {
   describe('`v-model` interface', () => {
     describe('when `currencyInput` changes', () => {
       it('emits an input event with the new value', () => {
-        wrapper.setData({ currencyInput: 25 });
-        expect(wrapper.emitted().input).toEqual([[25]]);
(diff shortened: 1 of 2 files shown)


tests/unit/components/Form/Currency.spec.js
```diff
@@ -62,8 +62,8 @@ describe('components/Form/Currency', () => {
   describe('`v-model` interface', () => {
     describe('when `currencyInput` changes', () => {
       it('emits an input event with the new value', () => {
-        wrapper.setData({ currencyInput: 25 });
-        expect(wrapper.emitted().input).toEqual([[25]]);
```
ollietreend: It looks like these old tests were still passing because the validator is on the prop (i.e. an incoming value). But you were changing an internal value and emitting, so it was an outgoing value. So the value wasn't being checked by the validator. 🙂

PR #58, opened 2019-09-16
Add Show Overview page

alexbokii:
**Change log:**

1) Add Overview page
2) Add classes into CSS for sections with grey and blue background

### DIFF
--- a/src/views/Exercises/Show/Overview.vue
+++ b/src/views/Exercises/Show/Overview.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div class="govuk-grid-column-one-half">
-      <div class="background-grey govuk-!-padding-4 govuk-!-margin-bottom-3">
+      <div class="background-light-grey govuk-!-padding-4 govuk-!-margin-bottom-3">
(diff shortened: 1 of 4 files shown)


src/views/Exercises/Show/Overview.vue
```diff
@@ -1,11 +1,78 @@
...
+  components: {
+    Timeline,
+  },
+  computed: {
+    exercise() {
+      return this.$store.state.exerciseDocument.record;
+    },
+    timeline() {
+      return [
+        { 'entry': 'Open for applications', 'date': this.$store.state.exerciseDocument.record.applicationOpenDate },
```
ollietreend: These lines seem quite long, and I think shortening them could help code readability here.

It's worth knowing that computed properties can make use of other computed properties. So, the `timeline` property could use the `exercise` property.

For example:

```js
{ entry: 'Open for applications', date: this.exercise.applicationOpenDate }
```

Alternatively, you could use a local variable to shorten the others:

```js
const exercise = this.$store.state.exerciseDocument.record;
// and then date becomes `exercise.applicationOpenDate`
```

PR #59, opened 2019-09-17
Checkbox Bug

LaurenQurashi:
This mighty bug 🐛 was causing our checkboxes to be plagued with a lack of flexibility. 

- Initially, the CheckboxGroup would have a lil fit if you initialised it with anything other than an empty array. To amend this, I've added a created method to the CheckboxGroup component that will emit it's value to be an empty array if it's set to anything that's not an array. This way, you can initialise it to whatever you fancy but at the end of the day, he'll be a big bad empty array, ready for some input. 

- Next issue was that the CheckboxItems were getting all picky about the CheckboxGroup value being set to null, but this was because they were checking the value of the CheckboxGroup before it had a chance to change itself to an empty array. To stop them being so damn hasty, we've added a v-if to the CheckboxGroup's item slot, to only render if the value is an array. 

This seemed to fix it. 

- Code has been linted.
- No additional tests have been written, not have originals been refactored, however they still pass. 
- Shortlisting.vue has been edited in order to test this new null acceptance. 

🌻

### DIFF
--- a/tests/unit/components/Form/CheckboxGroup.spec.js
+++ b/tests/unit/components/Form/CheckboxGroup.spec.js
@@ -219,22 +219,6 @@ describe('components/Form/CheckboxGroup', () => {
       });
     });
 
-    describe('created hook', () => {
-      describe('if value is an array', () => {
-        it('does not call emit', ()=> {
-          let array = [];
-          let wrapper = createTestSubject({ value: array });
-          expect(wrapper.emitted().input).not.toBeTruthy();
-        });
-      });
-      describe('if value is not an array', () => {
-        it('emits the initial empty array value', ()=> {
-          let wrapper = createTestSubject({ value: undefined });
-          expect(wrapper.emitted().input).toBeTruthy();
-        });
-      });
-    });
-
     describe('`.govuk-checkboxes` slot container', () => {
       describe('if value is an array ', () => {
         let slotContainer;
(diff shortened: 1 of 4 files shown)


tests/unit/components/Form/CheckboxGroup.spec.js
```diff
@@ -219,24 +219,61 @@ describe('components/Form/CheckboxGroup', () => {
       });
     });
 
-    describe('`.govuk-checkboxes` slot container', () => {
-      let slotContainer;
-      beforeEach(() => {
-        subject = createTestSubject();
-        slotContainer = subject.find('.govuk-checkboxes');
+    describe('created hook', () => {
```
ollietreend: Let's move the 'created hook' tests into a separate `describe()` block. They don't really fit inside the 'template' block, because they don't describe functionality of the `<template>`.

I tend to structure tests like this:

```js
describe('@/path/to/Component', () => {
  describe('template', () => {
    // ...
  });

  describe('lifecycle hooks', () => {
    describe('created', () => {
      // ...
    });
  });
});
```

Although I can see that `RepeatableFields.spec.js` doesn't nest the 'created' block inside a 'lifecycle hooks' block. In this case it doesn't matter hugely since this is the only lifecycle hook we're testing. So do whichever you prefer – so long as it's not nested inside the 'template' block.

PR #62, opened 2019-09-17
Exercise show eligibility page

ollietreend:
[Trello link](https://trello.com/c/kThXxdMk/116-details-page-eligibility)

### Things to note

- I've changed some of the inputs on the eligibility edit page as part of this work. Mainly that was to change String `"true"` and `"false"` values to Boolean `true` and `false`.
- The way we handle 'other' (please specify) fields is probably something which we could extract out into a helper method or component. It seems quite common.

### DIFF
--- a/src/views/Exercises/Edit/Eligibility.vue
+++ b/src/views/Exercises/Edit/Eligibility.vue
@@ -44,11 +44,11 @@
           hint="This lets appropriate candidates apply, even if they do not have the right legal experience."
         >
           <RadioItem
-            value="true"
+            :value="true"
             label="Yes"
           />
           <RadioItem
-            value="false"
+            :value="false"
             label="No"
           />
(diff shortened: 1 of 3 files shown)


src/views/Exercises/Edit/Eligibility.vue
```diff
@@ -184,13 +184,13 @@ export default {
 
     return {
       exercise: {
-        postQualificationExperience: exercise.postQualificationExperience || [],
+        postQualificationExperience: exercise.postQualificationExperience || null,
         otherYears: exercise.otherYears || null,
-        schedule2DApply: exercise.schedule2DApply || null,
+        schedule2DApply: exercise.schedule2DApply,
```
alexbokii: No fallback value for schedule2DApply @ollietreend ?

PR #63, opened 2019-09-17
Add Timeline show page

alexbokii:
**Change log**

Add Timeline show page

### DIFF
--- a/src/helpers/formatDate.js
+++ b/src/helpers/formatDate.js
@@ -1,0 +1,17 @@
+const formatDate = (date, type) => {
+  if(!(date instanceof Date)) {
+    return;
+  }
+
+  let month = date.toLocaleString('default', { month: 'long' });
+  let time = date.toLocaleString('en-US', { hour: 'numeric', minute: 'numeric', hour12: true });
+
+  if(type && type === 'time') {
+    return `${date.getDate()} ${month} ${date.getFullYear()} ${time}`;
+  }
+
+  return `${date.getDate()} ${month} ${date.getFullYear()}`;
+};
+
+export default formatDate;
+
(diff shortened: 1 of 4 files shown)


src/helpers/formatDate.js
```diff
@@ -0,0 +1,21 @@
+const formatDate = (date, type) => {
+  if(!(date instanceof Date)) {
+    return;
+  }
+
+  if(type && type === 'time') {
+    let time = date.toLocaleString('en-US', { hour: 'numeric', minute: 'numeric', hour12: true });
```
LaurenQurashi: Is there a British version for this? (look at me adding input 😎 )

src/helpers/formatDate.js
```diff
@@ -0,0 +1,21 @@
...
+    return `${time}`;
```
ollietreend: It looks like we don't really need to use a template literal string here.

Could it be simplified to just one line?

```js
return date.toLocaleString(...)
```

src/helpers/formatDate.js
alexbokii: Good find @LaurenQurashi!
There is, "en-GB". I'll update the PR :smiley:

src/helpers/formatDate.js
```diff
@@ -0,0 +1,21 @@
+const formatDate = (date, type) => {
+  if(!(date instanceof Date)) {
+    return;
```
ollietreend: This feels like a case where we could be `throw`ing an `Error` – so that as developers, it's obvious in our `console.error` output if we're using the method incorrectly.

We could have something like:

```js
throw "Supplied date must be a Date object";
```

PR #67, opened 2019-09-20
Add getter to get clone of the state object

alexbokii:
**Change log:**

To prevent modifying a record in the state we've added new getter that returns a clone of the state instead of an actual state object.

This change is implemented for add Edit and Show pages

### DIFF
--- a/tests/unit/store/exercise/document.spec.js
+++ b/tests/unit/store/exercise/document.spec.js
@@ -163,5 +163,28 @@ describe('store/exercise/single', () => {
         expect(exerciseDocument.getters.id(state)).toBe('abc123');
       });
     });
+
+    describe('exerciseData', () => {
+      it('returns a function', () => {
+        const state = {
+          record: {},
+        };
+        expect(exerciseDocument.getters.exerciseData(state)).toBeFunction();
+      });
+
+      it('returns a clone of the record data (rather than a reference to the state object)', () => {
+        const state = {
+          record: {
+            futureStart: 123, 
+            hmctsWelshGovLead: 'Test Name',
+            name: 'Test Name',
+          },
+        };
+
+        const recordObject = exerciseDocument.getters.exerciseData(state)();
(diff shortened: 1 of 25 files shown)


tests/unit/store/exercise/document.spec.js
```diff
@@ -163,5 +163,28 @@ describe('store/exercise/single', () => {
         expect(exerciseDocument.getters.id(state)).toBe('abc123');
       });
     });
+
+    describe('exerciseData', () => {
```
ollietreend: Small thought, but it might be a nice touch. How about putting `()` at the end of this getter name in the describe block? e.g. `describe('exerciseData()')`, or simply `data()` if renamed.

The reason I suggest this is it'd highlight that this isn't a 'normal' getter but a dynamic getter/function.

tests/unit/views/Exercises/Edit/Contacts.spec.js
```diff
@@ -10,6 +14,7 @@ const mockStore = {
   },
   getters: {
     'exerciseCreateJourney/nextPage': { name: 'mock-next-page' },
+    'exerciseDocument/exerciseData': () => () => exercise,
```
ollietreend: I think this getter should only be `() => exercise`, not a nested function.

Since this is only a mocked store, not a real Vuex store, we've set the getters to be the calculated value that Vuex would return for each getter.

This is why the `nextPage` getter is an object (not a getter function which needs Vuex to execute it to return an object). Likewise, the return value of `exerciseData` should be the function which returns the exercise, not a function which returns a function which returns the exercise.

Not sure if I've explained that very well! But this is all down to the fact that we're not really using Vuex, we're just mocking its API. And as far as our components are concerned, a getter should be a calculated value, not the function to generate that value.

tests/unit/store/exercise/document.spec.js
alexbokii: 👍 done!

PR #69, opened 2019-09-23
Textarea Input Component

LaurenQurashi:
Using the TextField component as a guide, I've created a TextareaInput component, that by default allows 5 rows of text. 

- Tests have been created and are passing, 
- Code has been linted,
- TextareaInput component has been added to the Vacancy-edit and Eligibility-edit pages to replace the hardcoded versions that were implemented on those pages beforehand. 

🍁

### DIFF
--- a/tests/unit/components/Form/TextareaInput.spec.js
+++ b/tests/unit/components/Form/TextareaInput.spec.js
@@ -52,16 +52,20 @@ describe('components/Form/TextareaInput', () => {
         expect(wrapper.find('label').attributes().for).toBe('my_unique_key');
       });
 
-      it('sets <input> `id` attribute', () => {
+      it('sets <textarea> `id` attribute', () => {
         wrapper.setProps({ id: 'my_unique_key' });
         expect(wrapper.find('textarea').attributes().id).toBe('my_unique_key');
       });
     });
 
     describe('rows', () => {
-      it('sets the rows to the value of the `rows` prop', () => {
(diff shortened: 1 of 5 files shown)


tests/unit/components/Form/TextareaInput.spec.js
```diff
@@ -0,0 +1,84 @@
...
+      });
+    });
+
+    describe('id', () => {
+      it('sets <label> `for` attribute', () => {
+        wrapper.setProps({ id: 'my_unique_key' });
+        expect(wrapper.find('label').attributes().for).toBe('my_unique_key');
+      });
+
+      it('sets <input> `id` attribute', () => {
```
ollietreend: This should read:

```
sets <textarea> `id` attribute
```

Not `<input>`

tests/unit/components/Form/TextareaInput.spec.js
```diff
@@ -0,0 +1,84 @@
...
+        wrapper.setProps({ id: 'my_unique_key' });
+        expect(wrapper.find('textarea').attributes().id).toBe('my_unique_key');
+      });
+    });
+
+    describe('rows', () => {
+      it('sets the rows to the value of the `rows` prop', () => {
```
ollietreend: It'd be good to capture somewhere that it has a default fallback of 5 rows.

So maybe a test to say it "defaults to 5 rows when no prop value is set"?

And then test without running `wrapper.setProps()` and setting the `row` prop.

PR #72, opened 2019-09-24
Create error pages

alexbokii:
This PR adds two 404 pages:
- Main 404 ("Page is not found")
- Exercise 404 ("Exercises is not found")

Also, this PR changes the error message in LoadingMessage component to be "Something went wrong"

**Change log**
1) Create 404 pages
2) Redirect to Exercise 404 page if exercise data is null using Vur router replace() method

### DIFF
--- a/tests/unit/views/Exercises/Edit.spec.js
+++ b/tests/unit/views/Exercises/Edit.spec.js
@@ -102,9 +102,10 @@ describe('@/views/Exercises/Edit', () => {
 
   describe('methods', () => {
     describe('redirectToErrorPage', () => {
-      it('calls router replace method', () => {
+      it('calls router replace method with the name of error page', () => {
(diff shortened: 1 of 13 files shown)


tests/unit/views/Exercises/Edit.spec.js
```diff
@@ -94,4 +99,13 @@ describe('@/views/Exercises/Edit', () => {
       });
     });
   });
+
+  describe('methods', () => {
+    describe('redirectToErrorPage', () => {
+      it('calls router replace method', () => {
+        wrapper.vm.redirectToErrorPage();
+        expect(mockRouter.replace).toHaveBeenCalled();
```
ollietreend: Can we use the same test from `Show.vue` here?

To test that we `.replace` to the correct route.

PR #73, opened 2019-09-24
Back Link Component 

LaurenQurashi:
Created a component that holds our back link functionality. The back link uses the `back()` functionality to return to the previous page when clicked. The link uses styling from the govuk style sheet. Tests have been created for this component. 

Code has been linted and tests are passing. 

- Should this component have an id prop? I thought it would help identify it, just incase there were two backlinks on a page, however is this a likely scenario where you wouldn't be able to distinguish between the two links that use the `.govuk-back-link` styling? 

- I hard coded the go(-1) functionality into all of the Edit pages before we decided that it would be a component. To add this component to the pages is part of another ticket, so I'll rectify this when I pick that up next. 

🍁


src/components/BackLink.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <a
+    class="govuk-back-link"
+    @click="$router.back()"
+  >
+    Back
+  </a>
+</template>
+<style scoped >
```
ollietreend: In terms of code style in this file, it'd be nice to have an empty line between `</template>` and `<style>`.

Also, there's an extra space which isn't needed after `scoped` which would be nice to clean up. It could simply be:

```
<style scoped>
```

tests/unit/components/BackLink.spec.js
```diff
@@ -0,0 +1,30 @@
+import { shallowMount } from '@vue/test-utils';
+import BackLink from '@/components/BackLink';
+
+describe('components/BackLink', () => {
+  let wrapper;
+  const mockRouter = {
+    back: jest.fn(),
+  };
+  beforeEach(() => {
```
ollietreend: The spacing and indentation is a little borked here – let's:

- add a blank line before the `beforeEach()` line
- add a space before the `=` in line `wrapper= shallowMount()`
- fix the indentation of the `beforeEach()` block – I think it's a little off right now

tests/unit/components/BackLink.spec.js
```diff
@@ -0,0 +1,30 @@
...
+      $router: mockRouter,
+    },
+});
+  });
+
+  it('renders the component', () => {
+    expect(wrapper.exists()).toBe(true);
+  });
+
+  describe('functionality', () => {
```
ollietreend: I think we could clean up the test descriptions a little here.

We might not even need any `describe()` blocks for this to be honest. This is such a simple component that all the `it()` tests could probably just sit in the main top-level `describe()` block.

I think we want to test:

- it renders the component
- it's styled as a [GOV.UK Back link](https://design-system.service.gov.uk/components/back-link/) – i.e. it has class `.govuk-back-link` and text 'Back'
- it calls `$router.back()` when clicked

I know you're already testing the router functionality, but I think we could clean up the descriptions / names of the tests.

PR #89, opened 2019-10-30
Additional amendments to the Eligibility Information page - *Will cause merge conflicts*

LaurenQurashi:
This PR *also* contains amendments to the Eligibility Information page. This could cause a merge conflict, so treat it carefully. 

Things that have been completed:
- The 'Post Qualification Experience' question only renders if the role is _not_ 'non-legal'. (ie You will _not_ see the PQE question if typeOfExercise is equal to 'non-legal'.)

Tests have been written and are passing. 

Code has been linted.

### DIFF
--- a/src/views/Exercises/Edit/Eligibility.vue
+++ b/src/views/Exercises/Edit/Eligibility.vue
@@ -12,6 +12,7 @@
         </p>
 
         <RadioGroup
+          v-if="typeOfExercise !== 'non-legal'"
           id="post-qualification-experience"
           v-model="exercise.postQualificationExperience"
           label="Post-qualification experience (PQE)"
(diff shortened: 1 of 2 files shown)


src/views/Exercises/Edit/Eligibility.vue
```diff
@@ -12,6 +12,7 @@
         </p>
 
         <RadioGroup
+          v-if="typeOfExercise !== 'non-legal'"
```
warrensearle: This works fine as is  though sometimes can be worthwhile using a `computed` property (e.g. `v-if="isNonLegal"`) then if there are any changes to the logic there's one place to update.
Like:
`computed: {
  isNonLegal () {
    return this.typeOfExercise !== 'non-legal'
  }
}`

PR #181, opened 2020-01-15
ADMIN:  Add Question to "Vacancy Information" about Statutory Consultation

LaurenQurashi:
Please add this question between "Jurisdiction" fields and "Appointment type"

Is statutory consultation waived for this exercise?
Yes or No.

If Yes add text area with a title: "Explain why". 

- I've added this question to both edit and show pages, and I've added the new variables to the data sheet. Were you planning on adding yes or no to the look ups? I've left them out as other variables on the show page didn't use lookup for their yes/no rendering, but if you want me to add them in then I can?

### DIFF
--- a/src/views/Exercises/Edit/Vacancy.vue
+++ b/src/views/Exercises/Edit/Vacancy.vue
@@ -167,6 +167,27 @@
         </CheckboxGroup>
 
         <RadioGroup
+          id="is-statutory-consultation-waived"
+          v-model="exercise.statutoryConsultationWaived"
+          label="Is statutory consultation waived for this exercise?"
+        >
+          <RadioItem
+            :value="true"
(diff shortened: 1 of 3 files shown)


src/views/Exercises/Edit/Vacancy.vue
```diff
@@ -166,6 +166,27 @@
...
+          label="Is statutory consultation waived for this exercise?"
+        >
+          <RadioItem
+            :value="true"
+            label="Yes"
+          >
+            <TextareaInput
+              id="statutory-consultation-waived-details"
+              v-model="exercise.statutoryConsultationWaivedDetails"
+              label="Explain why."
```
warrensearle: Please make this a required field

PR #189, opened 2020-01-16
Header size

londonprojects:


londonprojects:
I made changed to the header to work in the full column.

Issue #360, opened 2020-03-24
Ability for admin to switch between restricted panel view and full application


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
Exercise teams need to see a modified application view with reference number OR the full application with all information. 

**Describe the solution you'd like** (goes in order of application except personal details at the top)

One view should include everything

One view should NOT include:

personal details,
equality and diversity
character information
location preferences
jurisdictional preferences
interview preferences (new ticket)
The information should be grouped by the pages the information is stored. i.e all judicial experience should be together in one section.

**Describe alternatives you've considered**
A download that the teams could amend if necessary.

**Additional context**
Ultimately this view should be exportable

Linked to ticket #258

Rebecca-mcknight1991:
@warrensearle @YaaL does this summarise the need. This is the ticket linked to #258. If so please let us know estimate and we will move to Ready to Play.

warrensearle:
Perfect, thanks @Rebecca-mcknight1991. Let's put this down as an estimated 2 story points for the following reasons:
- 41 + 2 = 43 story points, which is another prime number
- the sister story is a 5 so a total of 7 points feels about right
@YaaL that sound agreeable to you?

YaaL:
@Rebecca-mcknight1991 @warrensearle 
Looks fine to me.

PR #373, opened 2020-04-01
Application full and panel pack view

YaaL:
Full and panel pack view of the application page.

Includes #360


src/views/Exercises/Show/Application.vue
```diff
@@ -1,71 +1,73 @@
...
+        </router-link>
+
+        <router-link
+          class="govuk-button govuk-button--secondary"
+          :to="{ name: 'exercise-show-application-streamlined' }"
+        >
+          Panel pack view
+        </router-link>
+        <span
+          v-if="!streamlined"
```
warrensearle, line 31: The `streamlined` variable doesn't exist in this component. Needs to either be passed in as a prop of defined as a local. Perhaps look at the current url?

YaaL: Looking at current url is a bit hacky.
I moved the buttons to the child component which has the prop available.

PR #376, opened 2020-04-06
Update diversity report to use pre-calculated data

warrensearle:
- Include refresh button to re-calculate report


YaaL:
Note: generate reports for already closed exercises.

PR #396, opened 2020-04-20
Reports for eligibility and character issues

warrensearle:
- Include new reports
- Amend application back link to use BackLink component


src/filters.js
```diff
@@ -173,13 +175,15 @@ const lookup = (value) => {
...
       'pakistani': 'Pakistani',
+      'pqe-issue': 'Post-qualification experience',
       'practice-or-employment-as-lawyer': 'Practice or employment as a lawyer',
       'pre-launch': 'Pre launch',
       'prefer-not-to-say': 'Prefer not to say',
       'property-chamber': 'Property Chamber',
       'ranked-choice': 'Ranked choice',
       'read': 'Read',
       'ready': 'Ready for approval',
+      'rls-issue': 'Reasonable length of service',
```
YaaL, line 186: These strings are generic and usable in other context, not just for this report, so key could be generic too.

warrensearle: Good point. Have updated so these no longer include '-issue'.

src/views/Exercises/Show/Reports/CharacterIssues.vue
```diff
@@ -0,0 +1,248 @@
...
+                :to="{name: 'exercise-show-application', params: { applicationId: application.id }}"
+                class="govuk-link print-none"
+              >
+                View application
+              </RouterLink>
+            </div>
+          </div>
+
+          <div 
+            v-for="(issue, index) in application.processing.characterIssues"
```
YaaL, line 51: Minor niggle, but it would be nice to have the issues in the same order they appear in the application, not just alphabetically.

warrensearle: Sorted in back end

PR #425, opened 2020-05-07
Independent Assessments

YaaL:
Independent Assessments page.

Note: needs more tests, especially for `ActionButton`


src/helpers/date.js
```diff
@@ -0,0 +1,73 @@
+const isDate = (date) => date instanceof Date;
+
+const isDateInFuture = (date) => {
+  // @NOTE: this is a bit silly, we should save full date instead of hardcoding the time
```
warrensearle, line 4: Hey @YaaL could you update this comment to reference issue admin#388 which covers saving time in the database. Also check if we need to add additional information / issue so that we refine this function.

YaaL: Done.

src/views/Exercises/Show/IndependentAssessments.vue
```diff
@@ -101,128 +210,135 @@
...
+  },
+  created() {
+    this.$store.dispatch('applications/bind', { exerciseId: this.exercise.id, status: 'applied' });
+    this.$store.dispatch('assessments/bind', { exerciseId: this.exercise.id });
+  },
+  beforeRouteUpdate (to, from, next) {
+    this.$store.dispatch('applications/bind', { exerciseId: this.exercise.id, status: 'applied' });
+    this.$store.dispatch('assessments/bind', { exerciseId: this.exercise.id });
+    next();
+  },
```
warrensearle, line 327: Hey @YaaL do we need both of these methods? Should there be a destroyed to unbind assessments?

YaaL: I'm not 100% sure if we need, possibly only the `created()`
As I understand, unbinding is not strictly necessary

warrensearle: Ok, lets consider in the next iteration

PR #432, opened 2020-05-11
Diversity report and Navbar updates

tomlovesgithub:


src/views/Exercises/Show.vue
```diff
@@ -39,13 +39,14 @@
         </div>
       </div>
       <div class="govuk-grid-row">
-        <div class="govuk-grid-column-one-quarter print-none">
```
warrensearle, line 42: Why were the print classes removed?

tomlovesgithub: Opps! just removed them as part of a process of elimination.
They were one of few differences so it was changed in the hunt for the real perpetrator, can be added back in if its not harming anything!

src/views/Exercises/Show/Reports/Diversity.vue
```diff
@@ -1,369 +1,410 @@
...
+    <div class="govuk-grid-column-full">
+      <div class="moj-page-header-actions">
+        <div class="moj-page-header-actions__title">
+          <h2 class="govuk-heading-l">
+            Diversity
+          </h2>
+        </div>
+        <div
+          class="moj-page-header-actions__actions" 
+          style="float: right;"
```
warrensearle, line 15: Is there a class that can be used for this instead of inline style? e.g. previous code used `text-right`

tomlovesgithub: yeah, this is an issue, if the previous code works then we can use that. I think i must have removed the `text-right` code in error, and i couldn't find the class responsible for putting it in the right place in the prototype classes from moj/govuk, so this was my way of making it right in the wrong way so it wouldn't go unfixed, hope that makes sense!

src/views/Exercises/Show/Reports/Diversity.vue
```diff
@@ -411,18 +452,44 @@ export default {
...
+}
+.govuk-table__header, .govuk-table__cell {
+    vertical-align: middle;
+}
+.govuk-table__header, .govuk-table__cell {
+    vertical-align: middle;
+}
+.govuk-table__header, .govuk-table__cell {
+    vertical-align: middle;
+}
```
warrensearle, line 483: Looks like there's a bit of duplication here ;-)

tomlovesgithub: 🤦‍♂️this is a bug from the VS Live share thing, occasionally save conflicts between participants will make this duplicates appear and appear <sup>and appear<sup>and appear<sup><sup>and appear</sup></sup></sup></sup>

tomlovesgithub: will certainly check for it in the future

src/views/Exercises/Show/Reports/Diversity.vue
warrensearle: 👍

src/views/Exercises/Show/Reports/Diversity.vue
warrensearle: ha ha great, thank you 👍

Issue #449, opened 2020-05-15
A few tweaks to Independent Assessments


warrensearle:
A few updates to Independent Assessments, as follows:
- [x] Hide the list of 'assessors listed by candidates'
- [x] Show 'Start Assessments' button if we don't have `exercise.assessments.initialised > 0`
- [x] Show list of assessments if we have do have `exercise.assessments.initialised > 0`
- [x] List of assessments should include a 'Send to All' button

Issue #230, opened 2020-01-24
Admin: Download and print Independent Assessments


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
As an admin I need to be able to download independent assessments on the platform and print them in PDF or word.

**Describe the solution you'd like**
Independent Assessment page: button to download assessment for each candidate.

warrensearle:
This is blocked by apply#135

Rebecca-mcknight1991:
Add screen grab of where this goes. @enoranidi @ClaireTroughtonJAC @Rebecca-mcknight1991

PR #461, opened 2020-05-19
Tweaks to IAs

YaaL:
#449 updated to initialise and send assessment requests separately
 #230 Updated Download button


src/views/Exercises/Show/IndependentAssessments.vue
```diff
@@ -71,241 +86,129 @@
...
-                />
-              </div>
-              <div
-                class="moj-button-menu__wrapper"
-              >
-                <ActionButton
-                  class="moj-button-menu__item"
-                  @click="resendRequest(assessment.id)"
+                  <DownloadLink
+                    v-if="true || assessment.fileRef"
```
warrensearle, line 160: Hey @YaaL the `true` can be removed now, right?

YaaL: Ah, yes.

PR #465, opened 2020-05-20
Scaffolding for exercise stages and application statuses

warrensearle:
Initial views for all exercise stages and editing application status


lloback:
THis is ready to merge

Issue #474, opened 2020-05-26
Notifications list


warrensearle:
Any emails sent through [Notify](https://www.notifications.service.gov.uk/) are first queued in a `notifications` collection. Let's expose this collection in Admin app and allow super admins to start and stop processing the notifications queue.

**To do**
- [x] @warrensearle Include notifications list to show queued notifications
- [x] @warrensearle Include notifications list to show sent notifications
- [x] @tomlovesgithub update styling of notifications list to include tabs
- [x] @warrensearle Add start/stop button to notifications list, to start/stop the queue being processed
- [ ] @warrensearle Introduce 5 minute delay between notification creation and sending, to give super admins time to stop the queue

PR #476, opened 2020-05-26
#474 Notifications

warrensearle:
Include notifications queue.

Admins can:
- see what's about to be sent out
- see what's already been sent out

### DIFF
--- a/src/components/Page/TabsList.vue
+++ b/src/components/Page/TabsList.vue
@@ -1,0 +1,45 @@
+<template>
+  <div>
+    <div>
+      <ul
+        role="tablist"
+      >
+        <li 
+          v-for="tab in tabs"
+          :key="tab.name"
+        >
+          <a
+            :href="tab.href"
+            @click="showTab(tab)"
+          >
+            {{ tab.name }}
+          </a>
+        </li>
+      </ul>
+    </div>
+    <div class="tabs-details">
+      <slot />
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    tabs: {
+      type: Array,
+      required: true,
+    },
+    isActive: {
+      type: Boolean,
+      default: false,
+    },
+  },
+
+  methods: {
+    showTab(selectedTab) {
+      console.log(selectedTab);
+    },
+  },
+};
+</script>
--- a/src/components/Page/TabsList.vue
+++ b/src/components/Page/TabsList.vue
@@ -1,18 +1,24 @@
 <template>
-  <div>
+  <div
+    class="js-enabled govuk-tabs"
+    data-module="govuk-tabs"
+  >
     <div>
       <ul
-        role="tablist"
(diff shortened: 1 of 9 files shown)


src/components/Page/TabsList.vue
```diff
@@ -0,0 +1,56 @@
...
+    },
+    activeTab: {
+      type: String,
+      default: '',
+    },
+  },
+
+  methods: {
+    showTab(selectedTab) {
+      console.log(selectedTab);
```
tomlovesgithub, line 47: my bad, this probably needs removing

src/views/NotificationsList.vue
```diff
@@ -0,0 +1,193 @@
...
+    this.$store.dispatch('notifications/bindQueue');
+    this.$store.dispatch('notifications/bindSent');
+  },
+  methods: {
+    showTab(tab) {
+      this.currentView = tab;
+    },
+    startProcessing() {
+      this.processing = !this.processing;
+      console.log('start processing');
```
tomlovesgithub, line 185: does this need removing too?

src/views/NotificationsList.vue
```diff
@@ -0,0 +1,193 @@
...
+    },
+    stopProcessing() {
+      this.processing = !this.processing;
+      console.log('stop processing');
```
tomlovesgithub, line 189: ...and this

src/views/NotificationsList.vue
```diff
@@ -0,0 +1,193 @@
...
+<script>
+import TabsList from '@/components/Page/TabsList';
+
+export default {
+  components: {
+    TabsList,
+  },
+  data() {
+    return {
+      tabs: [
```
tomlovesgithub, line 144: I did want to ask if this is a desirable method of passing in the tab names, perhaps someone else could imagine a better naming convention that i have here _(:tabs="tabs")_

src/components/Page/TabsList.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <div
+    class="js-enabled govuk-tabs"
```
tomlovesgithub, line 3: I _had_ to put the 'js-enabled' class here in order for the tabs styling to work, this is going to need to be changed at some point

warrensearle: We will get this added globally when vue starts up. See #482

src/views/NotificationsList.vue
warrensearle: There's another iteration of the Notification list coming along (today) so this will be considered then

src/views/NotificationsList.vue
warrensearle: This is fine for now. It will be sorted in the next iteration today...and before we release

src/components/Page/TabsList.vue
warrensearle: Sure

Issue #451, opened 2020-05-18
 Connecting statuses and stages so the candidates appear in the correct list


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
As an admin when I select a status update I need to move the candidate to the correct stage. This will sometimes have a dependency for example if there is a sift but another stage set up after in the candidate record the journey will be different for the candidate. 

**Describe the solution you'd like**
The candidate to either remain in the current stage where they don't progress or move into the next one. 

**Additional context**
This spreadsheet outlines the stages, statuses, where the candidate should go and whether there are any dependencies. 

https://docs.google.com/spreadsheets/d/1S6qpdM5DchujfNT0v0wUfpnVijtwor8sTcu44oB8E9A/edit?usp=sharing

warrensearle:
This is mostly done.
We have a couple of TODOs in our stores which will need to be completed, across all stores:
- [ ] Display available statuses based upon exercise configuration
- [ ] Move to next stage based upon exercise configuration

lloback:
![image.png](https://images.zenhubusercontent.com/5eb93282d6984623e15443b7/a999dde3-34c5-49ba-a615-2d70e3b66a31)

This is the table for testing the scenarios

lloback:
This task is done.

![image.png](https://images.zenhubusercontent.com/5eb93282d6984623e15443b7/91651c8a-05dd-4457-a9f2-0e0487d12689)

We introduced a "Move to the next stage" on the REVIEW tab in order to bypass some limitations in choosing the next stage. This can be clarified and a new task raised for this.

PR #487, opened 2020-05-28
#451 Review Status moved to Shortlisted

lloback:
On this commit I am moving the Stage from REVIEW to SHORTLISTED.

### DIFF
--- a/src/store/stage/review.js
+++ b/src/store/stage/review.js
@@ -3,7 +3,7 @@ import firebase from '@firebase/app';
 import { firestore } from '@/firebase';
 import { firestoreAction } from 'vuexfire';
 import vuexfireSerialize from '@/helpers/vuexfireSerialize';
-import { EXERCISE_STAGE, APPLICATION_STATUS, SHORTLISTING } from '../../helpers/constants';
+import { STATUS, EXERCISE_STAGE, APPLICATION_STATUS, SHORTLISTING } from '../../helpers/constants';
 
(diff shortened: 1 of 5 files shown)


src/store/stage/review.js
```diff
@@ -79,11 +79,22 @@ export default {
...
       return unbindFirestoreRef('records');
     }),
-    updateStatus: async ( context, { applicationId, status } ) => {
-      // @TODO based on provided status, work out whether stage should also be updated
+    updateStatus: async ( context, { applicationId, status, nextStage } ) => {
+      let stageValue = EXERCISE_STAGE.REVIEW; // initial value: 'review'
+
+      // Withdrawn
+      if (status === APPLICATION_STATUS.NO_TEST_SUBMITTED || status === APPLICATION_STATUS.WITHDREW_APPLICATION) {
+        stageValue = STATUS.WITHDRAWN;
```
warrensearle, line 87: My understanding is that the Stage will stay the same, not change to 'Withdrawn'. The main Application document will be updated to 'Withdrawn' but the Application Record document will remain in the stage it was at when the candidate withdrew. The status will however update to 'Withdrawn'.

lloback: Don't work this late in the night. It is not good for you! I will pretend I didn't see this at 11pm and will answer you tomorrow morning. Cheers

lloback: removed the withdrawn code.

PR #561, opened 2020-06-05
Feature/455 EMP flag creation

tomlovesgithub:


src/views/Exercises/Stages/RecommendedEdit.vue
```diff
@@ -24,6 +24,39 @@
...
+
+    <CheckboxGroup
+      id="emp-edit-toggle"
+      v-model="editEMPFlag"
+      label="Equal Merit Provision"
+      hint=""
+      required
+      :messages="{
+        required: 'Please specify whether you\'d like to add more information'
+      }"
```
warrensearle, line 36: This field isn't required and doesn't need a custom message.

src/views/Exercises/Stages/RecommendedEdit.vue
```diff
@@ -24,6 +24,39 @@
...
+    >
+      <CheckboxItem
+        :value="true"
+        label="Update EMP"
+      >
+        <RadioGroup
+          id="emp-edit-input"
+          v-model="editEMPFlagStatus"
+          label=""
+          hint=""
```
warrensearle, line 46: This field is required (when `editEMPFlag` has been ticked)

src/views/Exercises/Stages/RecommendedEdit.vue
```diff
@@ -92,7 +131,7 @@ export default {
       if (this.itemsWithIssues() && this.newSelectedStatus === APPLICATION_STATUS.APPROVED_FOR_IMMEDIATE_APPOINTMENT){
         this.showWarning = true;
       } else {
-        await this.$store.dispatch('stageRecommended/updateStatus', { status: this.newSelectedStatus });
+        await this.$store.dispatch('stageRecommended/updateStatus', { status: this.newSelectedStatus, empVal: { shouldUpdate: this.editEMPFlag, newStatus: this.editEMPFlagStatus } });
```
warrensearle, line 134: This could be simplified.
If you provide the `empApplied` field then the store will include it in the database update. If you don't include the `empApplied` field then the store won't include it in the database update.

src/filters.js
```diff
@@ -78,6 +78,16 @@ const candidateHasIssues = row => {
...
+const candidateHasEMP = row => {
+  let flag = false;
+  const flags = row.flags;
+  if (flags.empApplied) {
+    flag = true;
+  }
+
+  return flag ? 'Yes' : 'No';
+};
+
```
warrensearle, line 90: We don't need a new filter here.
We already have `toYesNo` so can do something like the following for empApplied:
```
{{ applicationRecord.flags.empApplied | toYesNo }}
```

PR #565, opened 2020-06-08
Feature/453 move back

lloback:
Back button for:

 SELECTED
 RECOMMENDED
 HANDOVER


src/router.js
```diff
@@ -305,6 +326,15 @@ const router = new Router({
                 title: 'Exercise Stages | Selected list',
               },
             },
+            {
+              path: 'handover/back',
+              component: ExerciseStagesHandoverBack,
+              name: 'exercise-stages-handover-back',
+              meta: {
+                requiresAuth: true,
+                title: 'Exercise Stages | Selected Back',
```
warrensearle, line 335: Other titles use lower case 'back' instead of 'Back'. Make consistent

PR #568, opened 2020-06-09
EMP further updates

tomlovesgithub:


src/store/stage/recommended.js
```diff
@@ -34,7 +34,7 @@ export default {
     unbind: firestoreAction(({ unbindFirestoreRef }) => {
       return unbindFirestoreRef('records');
     }),
-    updateStatus: async ( context, { status, nextStage, empVal } ) => {
+    updateStatus: async ( context, { status, nextStage, empApplied } ) => {
       const moveToNextStage = nextStage !== EXERCISE_STAGE.RECOMMENDED;
       const data = {
```
warrensearle, line 39: This isn't consistent with the other stores which always have `status` in the data object... but also conditionally overwrite it.

Issue #206, opened 2020-01-21
Restrict size and type of files that can be uploaded


Rebecca-mcknight1991:
Right now, there are no file size limits or file type restrictions. 

The restrictions we would like for uploads in admin are:
- docx, txt, pdf

See below image for error messages to display:
![image](https://user-images.githubusercontent.com/56963348/72826160-3d429980-3c70-11ea-825d-7d4b0a4f985c.png)

ClaireTroughtonJAC:
Update @warrensearle the file types we would like to restrict to are:

docx, txt, pdf, pages, google docs,

londonprojects:
@warrensearle will break up story

warrensearle:
This issue has been added to the following repos:

- Admin admin#206
- Apply apply#330
- Assessments assessments#42
- Digital Platform digital-platform#327

crablab:
The extensions I'm including for now: 'pdf', 'docx', 'doc', 'odf', 'pages'

(ODF is a open source equivalent to the propitiatory Microsoft formats. Word can open it.)

PR #576, opened 2020-06-11
#206 Candidates Detail

lloback:
On this PR I am adding the Candidate Detail view - it takes data from 3 documents on the 'candidates' collection

### DIFF
--- a/src/components/Candidates/CharacterInformation.vue
+++ b/src/components/Candidates/CharacterInformation.vue
@@ -1,0 +1,152 @@
+<template>
+  <div
+    class="govuk-!-margin-top-9"
+    v-if="Object.keys(data).length > 0"
+  >
+    <h2 class="govuk-heading-l">
+      Character information
+    </h2>
+
+    <dl class="govuk-summary-list">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Has been cautioned or convicted of a criminal offence
+        </dt>
+        <dd
+          class="govuk-summary-list__value"
+        >
+          {{ data.criminalOffences | toYesNo }}
+          <EventRenderer
+            v-if="data.criminalOffences"
+            :events="data.criminalOffenceDetails"
+          />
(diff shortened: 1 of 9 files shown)


src/components/Candidates/CharacterInformation.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+  <div
+    class="govuk-!-margin-top-9"
+    v-if="Object.keys(data).length > 0"
```
warrensearle, line 4: We should simplify this so we don't have 'complicated' code in our templates.
Perhaps for now use a computed prop such as:
`hasData() { return  Object.keys(this.data).length > 0; }`
Then our template can use `v-if="hasData"`

src/views/Candidates/CandidatesView.vue
```diff
@@ -0,0 +1,100 @@
...
+    };
+  },
+  computed: {
+    candidateRecords() {
+      const localRecords = this.$store.state.candidates.records;
+      return localRecords;
+    },
+    personalDetails() {
+      const localDocs = this.$store.state.candidates.personalDetails;
+      return localDocs || {};
```
warrensearle, line 71: What happens if we simply return what's in our store rather than returning an empty object?
Would this mean we don't need the `Object.keys` check inside the personalDetails component?

src/components/Candidates/CharacterInformation.vue
lloback: done.

src/views/Candidates/CandidatesView.vue
lloback: I was trying to fix the Object / Array Conundrum. And there is also a NULL value. I tried different approaches and this was the only one that worked. If you want I can look just into this issue.

warrensearle: I don't quite understand why it doesn't work as expected but the pattern you've used does the trick. So no further investigation needed 👍

Issue #202, opened 2020-01-21
ADMIN: Exercise document uploads.


Rebecca-mcknight1991:
Changes to exercise download pages: 

Please can you "add another" upload facility for Pension Note(s), Competency Framework and Welsh Translation.  This is a temporary measure until we can combined two exercises into one more effectively.  

Please can you a new title and upload fields for another REPEATABLE upload.  Title "Pension documents".  

We may have more than one document so this must be repeatable.  

These need to display and be downloadable to the candidate on the exercise information page under "Related Content".  

Checklist

- [x] Add a new downloadable field for "Pensions Information"

- [x] Allow for unlimited "add another" options for Pensions information

- [x] Add unlimited "Competency framework" upload field

~~- [ ] Keep one independent assessment template and Candidate assessment template only per exercise~~
        ⮑ moved to new ticket #587

- [x] Add a new downloadable field for "Welsh translation"

- [x] Allow for unlimited "add another" options for "Welsh translation"

- [x] Storage names for the additional pension note(s) and competency framework(s), the database can have this type of name pension note 1 or competency framework 2, the welsh translation 1 or translation 2.

Rebecca-mcknight1991:
@ClaireTroughtonJAC need clarity on this. @Rebecca-mcknight1991 write up fully.

YaaL:
@ClaireTroughtonJAC This will be partially done with what I'm doing for #372 - would be nice to clean this up.

PR #588, opened 2020-06-15
#202 Exercise Document Downloads

crablab:
Add upload/download for: 
- Pensions Information
- Competency Framework
- Welsh Translation

### DIFF
--- a/src/views/Exercises/Show/Downloads.vue
+++ b/src/views/Exercises/Show/Downloads.vue
@@ -133,6 +133,28 @@
           </ul>
         </dd>
       </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Competency Framework
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <span v-if="!exercise.downloads.competencyFramework.length">
+            No files uploaded
+          </span>
+          <ul class="govuk-list">
+            <li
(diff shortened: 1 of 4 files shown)


src/views/Exercises/Show/Downloads.vue
```diff
@@ -111,6 +111,72 @@
           </ul>
         </dd>
       </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Pensions Information
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <span v-if="!exercise.downloads.pensionsInformation.length">
```
warrensearle, line 119: The v-if here may be problematic for existing exercises, which might not have `downloads` or `downloads.pensionInformation`.
Would suggest you use a computed method to safely work out whether to show `pensionsInformation` or not.

src/views/Exercises/Show/Downloads.vue
```diff
@@ -111,6 +111,72 @@
...
+            </li>
+          </ul>
+        </dd>
+      </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Competency Framework
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <span v-if="!exercise.downloads.competencyFramework.length">
```
warrensearle, line 141: Same as previous comment

PR #592, opened 2020-06-16
Feature/515 notes

lloback:
This PR is to add a NOTES tab at the top and the ability to add notes to a candidate.

PS: On this PR you are not able to edit or delete the note

### DIFF
--- a/src/components/Candidates/Notes.vue
+++ b/src/components/Candidates/Notes.vue
@@ -4,7 +4,7 @@
   > 
     <NotesNew
       v-if="notesNew"
-      :noteId="notesId"
+      :note-id="notesId"
       @createdNote="createdNote"
     />
     <div v-if="!notesNew">
--- a/src/components/Candidates/Notes.vue
+++ b/src/components/Candidates/Notes.vue
@@ -1,0 +1,63 @@
+<template>
+  <div
+    class="govuk-!-margin-top-9 "
+  > 
+    <NotesNew
+      v-if="notesNew"
+      :noteId="notesId"
+      @createdNote="createdNote"
+    />
+    <div v-if="!notesNew">
+      <h2
+        class="govuk-heading-l"
+      >
+        Notes about the Candidate
+      </h2>
+
+      <button 
+        v-if="!notesNew"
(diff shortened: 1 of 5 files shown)


src/components/Candidates/NotesList.vue
```diff
@@ -0,0 +1,75 @@
...
+    :data="notesList"
+    :columns="[
+      { title: 'Date' },
+      { title: 'Note' },
+    ]"
+  >
+    <template #row="{row}">
+      <TableCell>
+        <RouterLink
+          :to="{ name: 'candidates-view', params: { id: row.id } }"
```
warrensearle, line 13: Not sure this link should go to the candidate-view. I guess it will link to edit the note, eventually? Otherwise could be removed

src/components/Candidates/Notes.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <div
+    class="govuk-!-margin-top-9 "
+  > 
+    <NotesNew
+      v-if="notesNew"
+      :noteId="notesId"
```
warrensearle, line 7: Need to use kebab case here `:note-id`

src/store/candidates.js
```diff
@@ -35,11 +35,55 @@ export default {
       const ref = collection.doc(`${id}/documents/personalDetails`);
       await ref.update(data);
     },
+    bindNotes: firestoreAction(async ({ bindFirestoreRef }, { candidateId, id }) => {
```
warrensearle, line 38: We may want to consider Notes being it's own store, as it's a top-level collection

src/components/Candidates/NotesList.vue
lloback: done.

src/components/Candidates/Notes.vue
lloback: Done.

src/store/candidates.js
lloback: Can I do it after this PR, when I am adding the DELETE / EDIT?

PR #601, opened 2020-06-16
Allow formatEstimatedDate to return TBC if no date providied

tomlovesgithub:


src/filters.js
```diff
@@ -19,15 +19,18 @@ const formatDate = (value, type) => {
...
+  if(value){
+    if (value instanceof Date) {
+      return formatDate(value, 'month');
+    }
+    const parts = value.split('-');
+    if (!parts[2]) {
+      return formatDate(value, 'month');
+    }
+    return formatDate(value);
+  } else {
```
warrensearle, line 31: Hey @tomlovesgithub where did this requirement come from? Are you sure this filter should return 'TBC' rather than leaving the value as it is?

PR #613, opened 2020-06-18
Feature/508 Initilize Sentry with env and release properties

### LINKED ISSUE #608: Add Sentry environments based on Node runtime environment
**Problem** 

In Sentry we can't easily filter between issues affecting Staging and those on Production, as they are all under "admin". 

**Solution** 

The Sentry SDK allows different environments to be [configured](https://docs.sentry.io/enriching-error-data/environments/). 

We could use `process.env.NODE_ENV` to create different admin Sentry environments (eg. `admin - production`), which will then show up in the Sentry UI and allow easy filtering. 

We could also provide the release version number ([docs](https://docs.sentry.io/workflow/releases/?platform=javascript)), which is taken from `process.env.npm_package_version`

crablab:
Uses environment variables for release and running environment to initialize Sentry, to allow better filtering of errors in the web UI.


crablab:
@warrensearle I note that the DSN isn't passed in via an environment variable. Whilst I don't think it is super secret (right?!), I think maybe we could open another ticket to push that out to .env?

src/main.js
```diff
@@ -8,12 +8,12 @@ import CKEditor from '@ckeditor/ckeditor5-vue';
 import * as Sentry from '@sentry/browser';
 import * as Integrations from '@sentry/integrations';
 
-if (process.env.NODE_ENV !== 'development') {
-  Sentry.init({
-    dsn: 'https://ab99abfef6294bc5b564e635d7b7cb4b@sentry.io/1792541',
-    integrations: [new Integrations.Vue({ Vue, attachProps: true })],
-  });
-}
+Sentry.init({
```
warrensearle, line 11: Remember to exclude sentry when we're in development mode - or at least when we're running code locally.

crablab: Apologies! I had a recollection that we were going to remove that, but I will add it back :+1:

PR #615, opened 2020-06-18
Feature/600 Add existence check for .review and .shortlisted on review dropdown

crablab:
Adds an existence check for the two options on the review dropdown, to prevent errors when either/or are undefined (ie. no candidate at that stage to review).


src/views/Exercises/Tasks/IndependentAssessments.vue
```diff
@@ -260,10 +260,18 @@ export default {
       return this.exercise.assessments && this.exercise.assessments.sent;
     },
     hasApplicationRecordsReview(){
-      return this.exercise.applicationRecords.review ? true:false;
+      if(this.exercise && this.exercise.applicationRecords && this.exercise.applicationRecords.review){
+        return false;
```
warrensearle, line 264: Think this should return true, right? :)

crablab: :facepalm:

PR #620, opened 2020-06-18
Remove redundant Details.vue

crablab:
This appeared to be a static template with dummy data. I couldn't see that it was referenced in router.js so I assume it is safe to remove? 

I haven't created a ticket, as this is an ad-hoc change.


lloback:
I think it is best practice to create a ticket - we need to discuss with @warrensearle  and @YaaL how do we want to handle, maybe calling `TECH TASK: delete unusued files` or something like that, and not sending it to @Rebecca-mcknight1991 or @ClaireTroughtonJAC to test. We need to test it ourselves.

This particular PR, I will leave it to @warrensearle  or @YaaL to review.

crablab:
> I think it is best practice to create a ticket 

Agreed - that makes sense. 

> There are other redundant files too. We have an upcoming epic to review and tidy the code base so this PR could reference that issue

I've added this to that Epic, and I'll add a specific ticket in that Epic to remove redundant files :+1:

Issue #626, opened 2020-06-23
Edit details of independent Assessors on Exercise Application Page


lloback:
**Is your feature request related to a problem? Please describe.**
As an admin I want to be able to edit the name, email and telephone of the independent assessors.

**Describe the solution you'd like**
The system will allow for each individual field to be edited.

**Describe alternatives you've considered**
Alternative would be to have a new page with a new navigation and URL

**Additional context**
![image](https://user-images.githubusercontent.com/15261814/85434337-cb071580-b57d-11ea-9ce3-cae9534e403a.png)

PR #630, opened 2020-06-25
#626  New component for Editable Fields

lloback:
Possibilty of changing details of the Assessors
Added a new shared component EditableField.vue

### DIFF
--- a/src/components/EditableField.vue
+++ b/src/components/EditableField.vue
@@ -39,7 +39,7 @@
         class="govuk-button"
         @click="btnClickSubmit()"
       >
-        Submit
+        Save
       </button>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


src/components/EditableField.vue
```diff
@@ -0,0 +1,122 @@
...
+    >
+      <TextField 
+        :id="`editable-field-${id}`"
+        v-model="localField"
+      />
+      <button 
+        class="govuk-button"
+        @click="btnClickSubmit()"
+      >
+        Submit
```
YaaL, line 42: We prefer to use "Save" CTA

lloback: Changed to save

Issue #636, opened 2020-07-01
Stages lists:  Ability to see more than 50 candidates at a time


ClaireTroughtonJAC:
**Is your feature request related to a problem? Please describe.**
We have feedback from staff that it is difficult to update statuses when not all the candidates appear in the list.  The current view shows only 50 and until you update a status the other candidates do not appear.  

**Describe the solution you'd like**
I would either like a way of displaying all candidates or through the updating of statuses on each page with 50 on one page.  

**Describe alternatives you've considered**
Ultimately I want to sort columns and filter information - ie if you update 25 people to a passed status I can filter those by only showing those who had applied.  

**Additional context**

https://apply-admin.judicialappointments.digital/exercises/2k6Qo2JaesamXjhreHiq/stages/shortlisted

<img width="813" alt="Screenshot 2020-07-01 at 10 14 41" src="https://user-images.githubusercontent.com/27224574/86226553-bd581e00-bb83-11ea-8966-326b6808325e.png">

ClaireTroughtonJAC:
Comment from user 

Can you please extend the list of 50 candidates that we can see on the platform to the full list of candidates selected for interview. I need to change the status of one candidate who has withdrawn due.

crablab:
Just copying from my messages with @lloback: 

So for this one the 'easy' option seems to just be to increase the number of rows shown in the table. But that isn't a great option as it doesn't solve the underlying issue, doesn't scale and you end up with a huge page.
The better option would be to add some pagination on the table component, I think. 

We could take a parameter with the exerciseID in the Store to set the limit. Ideally we'd add a dropdown to set this limit on the frontend, but I think hardcoding it for now would be okay.

If you added forward/back buttons to the component you could then just adjust the indexes on the array that are shown, and skip forwards/back through said array. 

One thing to note is that we should make this change across all of the `stages` pages (Shortlisted, Selected, Review, Recommended, Handover) for consistency and because it will likely come up again anyway.

enoranidi:
Agree with @crablab that we should not add more rows unless there are several users who have this same issue. 

If there is a user need then we could try to increase the number of table rows to say 100 (or whatever 'number' is reasonable from the users perspective).

I have a pagination example in my prototype, but here is a screenshot to show how the user can go back and forth:
<img width="1084" alt="Screenshot 2020-07-01 at 10 36 31" src="https://user-images.githubusercontent.com/1918282/86228806-c3033300-bb86-11ea-8695-cae695432cc3.png">

[...]

PR #649, opened 2020-07-06
Feature/636: Table Pagination

crablab:
Adds pagination to the Table component and updates store for Stages to take a `limit` parameter set in the page. 

Screenshot from 2020-07-06 08-56-35

Caveats: 
- There is still a hard limit on how many items are loaded from the database (but it is now higher). We could add a dropdown in the frontend, but we should probably instead make paging dynamically load each page, improving performance and removing any arbitrary limit. 
- The "select all" checkbox literally selects all the data that has been loaded into the component (up to 500 rows). This is probably not what we want it to do, but I can't see it being used that often so we can probably live with it for now. We could add a "X rows selected", and simply by implementing the aforementioned data paging, you will get this effect for fee.

### DIFF
--- a/src/components/Page/Table/Table.vue
+++ b/src/components/Page/Table/Table.vue
@@ -80,34 +80,28 @@
         <li class="moj-pagination__item  moj-pagination__item--prev">
           <a
             class="moj-pagination__link"
-            href=""
+            @click.prevent="changePage('previous')"
           >Previous<span class="govuk-visually-hidden"> set of pages</span></a>
         </li> 
-        <li class="moj-pagination__item">
+        
+        <li
+          v-for="number in calculatePages"
+          :key="number"
+          scope="col"
+          class="moj-pagination__item"
+        >
           <a
             class="moj-pagination__link"
-            href="#"
-          >9</a>
+          >{{ number }}</a>
         </li> 
-        <li class="moj-pagination__item moj-pagination__item--active">
-          10
-        </li> 
-        <li class="moj-pagination__item">
-          <a
-            class="moj-pagination__link"
-            href="#"
-          >11</a>
-        </li> 
+
         <li class="moj-pagination__item  moj-pagination__item--next">
           <a
             class="moj-pagination__link"
-            href=""
+            @click.prevent="changePage('next')"
           >Next<span class="govuk-visually-hidden"> set of pages</span></a>
         </li>
       </ul>
-      <p class="moj-pagination__results">
-        Showing <b>10</b> to <b>20</b> of <b>30</b> results
-      </p>
     </nav>
   </div>
 </template>
@@ -140,13 +134,13 @@ export default {
     pageSize: {
       type: Number, 
       required: false, 
-      default: 2,
+      default: 25,
     },
-    pageNumber: {
-      type: Number,
-      required: false, 
-      default: 1,
-    },
+  },
+  data(){
+    return {
+      pageNumber: 1,
+    };
   },
   computed: {
     selectAll: {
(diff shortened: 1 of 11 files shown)


crablab:
Hi @warrensearle, 

Thanks for the comments :) 

>  All these additions/changes should be made in the sandbox, away from actual usage of the components.

Would it be possible to understand a bit more about: 
a) what this sandbox is? 
b) how to use it? 
c) why we wouldn't make the change directly to the component? 

> which should be covered by separate, specific tickets.

I'm afraid I don't really understand why we need to do this? It seems to be creating tickets for the sake of creating tickets. The component (in whatever form) was designed to solve a particular problem, identified in #636. 

The work is scoped to the "Stages" area of `admin`. It may be that we find this pattern useful elsewhere in the application, but I don't think that should stop us making the changes there initially. 

For example. the "Notifications" section isn't even using the table component itself. So I think adding these changes to areas beyond the initial scope is certainly something for other tickets, but changes to a small and well defined area are fine. This is especially true if the changes are actually to support using the component to deliver the business outcome. 

---

I think it's important to note this is a fairly business critical change, and there is a demand for getting something out there. This doesn't mean we shouldn't do it right the first time around (eg. unit tests), but it would be good to understand how important some of the suggestions are, which is difficult to do when there isn't a great deal of explanation behind some of the requested changes. 

To that end: 

> This should be re-organised so there's a separate pagination component, which our table component is then amended to make use of.

I can see now why this might be advantageous. I'm happy to look into doing it, but I would question where else we are likely to need to the pagination component _urgently_, and therefore how important it is that this is abstracted right now. 

Thanks :)

warrensearle:
Hey @crablab appologies if my original feedback was a little sparse. It was a follow-up on the conversation you had with @lloback this morning at stand-up. I shall elaborate a little bit here but also DM/call you.

So to bring some context to this, we have an existing Table component which is used across the whole app. It is already being used in the following locations and also being rolled out to pretty much every list page in the app:
- Candidates list
- Notes list
- Applications lists
- Stages lists
Changes we make to the Table component will effect all these current and future instances of it's use.

Therefore in changing or adding to 'global' components like this, one approach we've taken is to use a Sandbox page.
This is a simple page/environment where we control the data fed into the component we're working on and can see the results in isolation from the wider app. Let's go through this further in person.

The other approach that's usually good to take is new functionality like this should generally be optional. In this instance most uses of the Table component will indeed want paging but perhaps paging should only be turned on if the view using the component provides a prop to turn it on - e.g. the `pageSize` prop might be used to turn on paging `<Table page-size="50">`.

What you've done so far is indeed good and appreciated just I think there are three stages/tickets to the work:
1. This was based upon a user-reported issue where list sizes are limited to 50. The immediate and temporary fix for this is to remove/increase that limit so that users can do what they need to do. (this is not ideal but is ok as we know that we're not dealing with loads of data yet). See admin#636.
2. This highlights the need for adding pagination to our Table component. So we should feed into and prioritise the ticket we have for that. See admin#542
3. When our Table component supports pagination we should identify all the views that are using it and consider whether they all want pagination turned on, page size etc and if any supporting changes are needed such as updating our stores to support paging. See TBC.

My thoughts are that this PR should be for either 1) or 2) above rather than a mix of all three.

In terms of a Pagination component, it looks to me like that block of code (Table.vue, lines 69 - 107) could be encapsulated in its own component, along with supporting logic. Primarily this would keep Table.vue smaller and so easier to work with. In addition though we might want to display the pages links at the top of the table (or both top and bottom). Or maybe we'd use the same component outside of the Table, for example in an image gallery or document viewer.

Hope this helps to some extent. Let's discuss further in a call.

crablab:
Thanks @warrensearle  - this is really helpful :smiley: 

> This is a simple page/environment where we control the data fed into the component we're working on and can see the results in isolation from the wider app.

This makes sense, thanks, I think this should probably be separate from the actual service, and there is a wider discussion to be had about centralising our component libraries and abstracting them from the actual codebase, given the reuse across repositories, which would then allow an independent sandbox,

I don't think we should be afraid of larger PRs, however. Hopefully increasing test coverage will help here, but changing a single component in one place (regardless of re-use across the codebase) should not require several PRs, and indeed that makes it more difficult to see what has changed. With better coverage we should be able improve this, and therefore increase the pace of delivery. 

> The other approach that's usually good to take is new functionality like this should generally be optional.

So I agree this is something I should have thought about! If you note, there is a computed property for this (which currently does nothing). 

However, as you say "in this instance most uses of the Table component will indeed want paging", as mostly we will want paging enabled I think we should default to 'true', rather than disabling by default. I think that makes sense given the context in which we are using the component? 

> The immediate and temporary fix for this is to remove/increase that limit so that users can do what they need to do. 

I, personally, think we should prioritise a proper if not complete fix for this. Since it requires deployment anyway, we might as well try to get a better change into the next train as quickly as possible. 

We can be agile about this and get _something_ out there to address the user need. I think a huge page could be a regression in terms of usability, which is why I prioritised getting this out in some form. 

> When our Table component supports pagination we should identify all the views that are using it and consider whether they all want pagination turned on, page size etc and if any supporting changes are needed such as updating our stores to support paging

Sure. I think that belongs here, rather than in a separate PR though. I can understand why we might want to avoid expending a load of energy on something that won't pass code review, but the more fragmented the problem becomes, the more difficult to track the changes and the longer it takes to get it co-ordinated and done. 

This PR creates a component that we know we need to use in several Stages. If we don't want to roll out the change across all of the rest of the instances of the Table, we could make it disabled by default (as you suggest) for the moment, which would mean no change to the other instances. A follow on PR would then be to change it to the default behaviour, having confirmed there are no issues elsewhere - using the Stages instance as a "beta", if you will. 

I think we might also want to review the 1:1 relationship between PRs and Issues (ticket) - I'm not convinced that is especially advantageous, especially for small pieces of work that don't necessarily need a ticket. A PR is self documenting and we could be more agile here.  

> In terms of a Pagination component, it looks to me like that block of code (Table.vue, lines 69 - 107) could be encapsulated in its own component, along with supporting logic. Primarily this would keep Table.vue smaller and so easier to work with. In addition though we might want to display the pages links at the top of the table (or both top and bottom). Or maybe we'd use the same component outside of the Table, for example in an image gallery or document viewer.

Agreed, this is the way to go. It will also make unit testing easier. I'll get this done today.

PR #652, opened 2020-07-08
#604 Add existence and type check for value passed to computed prop

crablab:
Adds existence and type checking for the `val` parameter passed to the computed prop, to handle the case that a null value is passed.


src/views/Exercises/Edit/Summary.vue
```diff
@@ -95,6 +95,10 @@ export default {
         return this.parseDate(this.exercise.estimatedLaunchDate);
       },
       set(val) {
+        if(!val && val instanceof Date){
+          return;
+        }
+
```
warrensearle, line 101: Please double check this.

crablab: Yes, as discussed I'm not sure what I was thinking here... 

I suspect this is meant to be along the lines of: 

```
if (val == null || !(val instanceof Date)) {
   return;
}
``` 

I shall have a look and fix this.

crablab:
@warrensearle Ah, it was my understanding that this should not be possible. 

To confirm, we want to allow a null value to be saved? In which case, we just need to handle that case rather than trying to transform a null value into a date :+1:

crablab:
@ClaireTroughtonJAC Would you be able to just confirm the understanding here either way? 

Understanding: Once a launch date has been provided (eg. when the Exercise is created) it can be modified, but it cannot be set to `null` (ie. no launch date).

crablab:
Claire has confirmed this is correct :+1:

PR #667, opened 2020-07-14
PR cleanup

tomlovesgithub:


src/views/Exercises/Applications/Application.vue
```diff
@@ -782,65 +782,56 @@
...
                   <dt class="govuk-summary-list__key">
-                    Date completed pupillage
+                    Completed pupillage
                   </dt>
                   <dd class="govuk-summary-list__value">
                     <ul class="govuk-list">
-                      <li v-if="item.date">
-                        {{ item.date | formatDate }}
+                      <li>
+                        {{ !item.qualificationNotComplete | toYesNo }}
```
warrensearle, line 818: As this is inside a `v-if="item.qualificationNotComplete"` we could probably ditch the binding and just have "No" here. This would then be slightly simpler if we need to make future changes.
That said this block could possibly be combined with the block starting line 824

Issue #661, opened 2020-07-13
Candidates' full application not displaying on Application or Stages view


ClaireTroughtonJAC:
**Describe the bug**
For all exercises where Additional Selection Criteria has been added to the exercise it is not displaying in the application information 

From a look at the database the information is being captured in the application but is not being displayed in the review page on apply or on Admin in either the Full application view or the panel pack view.  

**To Reproduce**
1. Go to 'JAC0028.'
2. Click on 'Applied'
3. Click on a candidate name
4. Click on Applications tab
5. Then view application
6. Then look at the full information and panel views tab
7.  There is no Additional selection criteria displaying

**Expected behavior**
Both admin and panel members will need to see the information.  

**Screenshots**
![Screenshot 2020-07-13 at 10.45.56.png](https://images.zenhubusercontent.com/5e3c2b35e9bf7c0ea1c2859a/e9deac12-ec19-4289-8b5d-52e995fb4a59)
![Screenshot 2020-07-13 at 10.45.30.png](https://images.zenhubusercontent.com/5e3c2b35e9bf7c0ea1c2859a/7b10bb15-cf42-4161-acc7-17beefbc3e93)

crablab:
@ClaireTroughtonJAC Thanks Claire! :raised_hands:  

Please could I just clarify:

> For all exercises where Additional Selection Criteria has been added to the exercise it is not displaying in the application information

Have we confirmed this is the case? If so, are there further examples with different ASC options that this impacts? 

For a Legal exercise, which 028 is, what ASC apply? In the database this is currently recorded as `self-assessment-with-competencies`.

YaaL:
Currently the ASC are only displayed when one of the following assessment options has been selected:
* Statement of suitability
* Statement of eligibility
This is true both for apply and admin.

Exercise JAC00028 has _Self Assessment with competencies_ selected.

crablab:
This was my confusion too... 

I assume this means someone has changed the option half way through?

ClaireTroughtonJAC:
Thank @YaaL 
I think that is the case @crablab 

However for all exercises where ASC is chosen it is supposed to be displayed on apply on these pages

Statement of Suitability - Leadership exercise. E.g 001
CV - Leadership exercises E.g 001
Statement of Eligibility- e.g 011
Self Assessment - 028

It then should follow that the ASC displays on the full application form view and the panel packs

YaaL:
[...]

PR #668, opened 2020-07-14
#661 - Missing additional selection criteria in Application.vue

crablab:
This has been a bit confusing :sweat: 

The ASC should display regardless of whether SoS, CV etc. are required in addition. 

This change moves the ASC out of the SoE and SoS blocks so they always display when they exist. It also removes the SoE block which does not have a file upload, and therefore will be encompassed in the ASC block. 

This will show in the panel pack too.

### DIFF
--- a/src/views/Exercises/Applications/Application.vue
+++ b/src/views/Exercises/Applications/Application.vue
@@ -1478,36 +1478,6 @@
             </div>
 
             <div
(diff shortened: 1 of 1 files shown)


crablab:
@warrensearle This is the correct behaviour and has been confirmed with @ClaireTroughtonJAC :smile:

PR #674, opened 2020-07-14
669: try out pagination on ShortlistList

crablab:
Draft PR for comments and thoughts :pray:  

Several things to note about the implementation: 
- This does not use [Firestore pagination](https://firebase.google.com/docs/firestore/query-data/query-cursors) deliberately for performance reasons. In order to paginate properly, we need to know how many records there are in the first place (ie. how many pages). As Firestore has no mechanism for us to retrieve just a count of the query, we have to load all the records any way (just to count them). It is therefore undesirable to initially request all the data, then continue making requests to Firestore each time a user changes page for the page of data we want - having loaded all of it, we might as well slice it ourselves. 
- There are no changes to the Store. It's easier to make these changes in `ShortlistList` and I would argue it is more semantic. It does mean the slice function, for example. will be replicated. If we consolidate stores we should look at bringing this into the store, but it does add complexity (as the store then has to keep track of how many records it is going to load, before anything has asked it to load them, and also issues with caching the loaded data). If there are strong feelings, I am happy to look further into putting this into the store :+1: 

I would just note, that this would have been much easier with `COUNT()` :smile:

### DIFF
--- a/src/views/Exercises/Stages/HandoverList.vue
+++ b/src/views/Exercises/Stages/HandoverList.vue
@@ -27,7 +27,7 @@
       </div>
       <Table 
         data-key="id"
-        :data="applicationRecords"
+        :data="getPaginated"
         :columns="[
           { title: 'Reference number' },
           { title: 'Name' },
@@ -58,6 +58,10 @@
           <TableCell>{{ row.flags.empApplied | toYesNo }}</TableCell>
         </template>
       </Table>   
+    <Pagination 
+        :high-index="numberOfPages"
+        @pageChanged="updatePage($event)"
+      />
     </form>
   </div>
 </template>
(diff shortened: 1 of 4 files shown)


crablab:
(Just to note, if we ever wanted to do proper pagination with the database only loading a single page, along with stuff like full text search, we basically have to have a proper DBMS, as we need to be able to count a very specific query without loading the entire result set, which won't have previously been cached)

warrensearle:
So the compromise here is we should display Next and Previous buttons to the user and **not** page numbers.
This is quick and simple to implement.
The question is whether our users need page numbers.
I'd argue perhaps not.

warrensearle:
Likewise we could use continuous scroll instead of pagination.

crablab:
We can adapt the component to not have page numbers, but I'd argue that they likely would be necessary, especially when dealing with large lists. 

It's much easier to remember that candidate X is on page 2 and candidate Y on page 22, than remember where they are in an infinite scroll (people don't seem to use find in page :see_no_evil:) 

From @llobacks's comment, I'm taking that we want to implement it like this for now? I think we could work on fixing the inefficiencies later when they become an issue and something breaks, especially if we're going to be doing a major database change. 

If so, I'll add this to the other stages?

Issue #595, opened 2020-06-16
Application view displays 'Invalid Date' for created


warrensearle:
**Expected**
Show the date

**Actual**
Shows 'Invalid Date'
See screen grab
![Screenshot 2020-06-16 at 12.08.07.png](https://images.zenhubusercontent.com/5e412648a39708a606c962c9/a3a237b0-0fde-4b95-8308-3f53b4f830b3)

**Example:**
https://admin.judicialappointments.digital/exercises/EMZEoUTXaqhUBvvrPEyv/application/6cbBRLaGsz12g8BexXar

crablab:
@warrensearle Apologies, looks like I forgot to follow up on this one. 

In some cases there is a data issue which means there isn't a date to display. So the question is really what should we display when we don't have a date to display? 

I would suggest something like "Unknown". It only impacts a small number of past records.

warrensearle:
Hey @crablab I'm happy with your proposed solution. Let's go with it.
@ClaireTroughtonJAC and @Rebecca-mcknight1991 can suggest a text change, if necessary, when reviewing this fix.

crablab:
Awesome! PR coming :soon:

YaaL:
This now works as expected.

There is however an underlying issue with the data - `createdAt` is stored as a Unix timestamp instead of Firebase Timestamp object.

crablab:
@warrensearle There is a separate issue for this now open: https://github.com/jac-uk/digital-platform/issues/331

PR #713, opened 2020-07-29
#595 - Add null check for application submitted/created dates

crablab:
This adds a couple of turnery checks to see if `application.createdAt` and `application.appliedAt` are null. 

The behaviour is now:
- If there is a valid date, the date is formatted and displayed
- If there is no date, we display "unknown" (new behaviour)
- If there is data in the field but it is not a valid date, "Invalid Date" will be shown (I think this is expected) 

The reason this took so long is Vue weirdness in it's overloading of bitwise-OR: https://github.com/vuejs/vue/issues/5449#issuecomment-294337677

Very much a TIL moment.

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -3,6 +3,12 @@ import { APPLICATION_STATUS } from '@/helpers/constants';
 const formatDate = (value, type) => {
   if (value) {
     const objDate = new Date(Date.parse(value));
+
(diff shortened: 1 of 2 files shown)


crablab:
The failed CI seems to be a result of linting errors introduced by another PR :grimacing:

PR #716, opened 2020-07-29
#206 - Restrict file uploads to certain extensions

crablab:
Changes the `FileUpload` component in admin to restrict the type of extensions that can be uploaded. This automatically works for anything that imports `FileUpload` eg. `MultiFileUpload`. 

Includes tests for the new functionality. 

Notes: 
- This is not enforced on the backend/in the bucket. (https://github.com/jac-uk/digital-platform/issues/327)
- This only checks the declared extension, not what the file content actually is 
- This only affects admin. I'll replicate the changes once approved (please monorepo soon!) 

Please let me know your thoughts :smile:

### DIFF
--- a/src/components/Form/FileUpload.vue
+++ b/src/components/Form/FileUpload.vue
@@ -131,11 +131,15 @@ export default {
     validFileExtension(originalName){
       const parts = originalName.split('.');
 
-      if (parts.length < 2 || !this.acceptableExtensions.includes(parts[parts.length - 1])){
+      if (parts.length < 2){
         return false;
       }
 
-      return true;
+      if (this.acceptableExtensions.includes(parts.pop())){
+        return true;
+      }
+
+      return false;
     },
     resetFile() {
       this.$refs.file = null;
@@ -146,7 +150,8 @@ export default {
       if (!file) {
         this.setError('File upload failed, please try again');
         return false;
-      } else if (!this.validFileExtension(file.name)) {
+      } 
+      if (!this.validFileExtension(file.name)) {
         this.setError(`Invalid file type. Choose from: ${this.acceptableExtensions}`);
         return false;
       }
--- a/src/components/Form/FileUpload.vue
+++ b/src/components/Form/FileUpload.vue
@@ -85,6 +85,7 @@ export default {
       file: '',
       isReplacing: false,
       isUploading: false,
+      acceptableExtensions: ['pdf', 'docx', 'doc', 'odf', 'pages'],
     };
   },
   computed: {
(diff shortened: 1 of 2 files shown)


tests/unit/components/Form/FileUpload.spec.js
```diff
@@ -257,6 +262,78 @@ describe('components/Form/FileUpload', () => {
...
+      });
+
+      it('accepts .doc files', () => {
+        const setName = 'filename';
+        wrapper.setProps({ name: setName });
+
+        expect(wrapper.vm.validFileExtension('original name.doc')).toBeTruthy();
+      });
+
+      it('accepts .odf files', () => {
```
YaaL, line 294: Duplicate

tests/unit/components/Form/FileUpload.spec.js
```diff
@@ -257,6 +262,78 @@ describe('components/Form/FileUpload', () => {
...
+      });
+
+      it('accepts .pages files', () => {
+        const setName = 'filename';
+        wrapper.setProps({ name: setName });
+
+        expect(wrapper.vm.validFileExtension('original name.pages')).toBeTruthy();
+      });
+
+      it('accepts layers.of.indirection.docx files', () => {
```
YaaL, line 308: I'd call it 'accepts files with multiple extensions if last one is valid'

tests/unit/components/Form/FileUpload.spec.js
```diff
@@ -257,6 +262,78 @@ describe('components/Form/FileUpload', () => {
...
+        const setName = 'filename';
+        wrapper.setProps({ name: setName });
+
+        expect(wrapper.vm.validFileExtension('original name.layers.of.indirection.pdf')).toBeTruthy();
+      });
+      
+      it('rejects .png files', () => {
```
YaaL, line 315: Or 'rejects files with unrecognised extension'

src/components/Form/FileUpload.vue
```diff
@@ -127,16 +128,27 @@ export default {
...
       this.$refs.file = null;
       this.isUploading = false;
     },
     async upload(file) {
       // @todo return more useful error messages
-
       if (!file) {
         this.setError('File upload failed, please try again');
         return false;
+      } else if (!this.validFileExtension(file.name)) {
```
YaaL, line 149: Can be just next `if`, doesn't need to be `else if` since we're returning false in the previous block.

src/components/Form/FileUpload.vue
```diff
@@ -127,16 +128,27 @@ export default {
 
       return [this.name, parts.pop()].join('.');
     },
+    validFileExtension(originalName){
+      const parts = originalName.split('.');
+
+      if (parts.length < 2 || !this.acceptableExtensions.includes(parts[parts.length - 1])){
```
YaaL, line 134: Can use `parts.pop()` here

YaaL, line 134: Style note: you can split this into two separate ifs for readibility since both return false.

crablab:
Thanks @YaaL! Will address all these tomorrow :sparkles:

Issue #711, opened 2020-07-28
Flag for team if IA is submitted late


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when Assessors upload Independent Assessments late and I do not know. 

**Describe the solution you'd like**
The assessor can put an assessment in late but it gets flagged on admin as past the due date entered by the team. The team get a notification and they can then choose to allow or deny the assessment - this would not appear to the assessor. Most extensions are agreed in advance so the team would go in check and then allow the late submission. 

This would **not** apply to assessors wanting to change details, once an assessment has gone in and the deadline has passed they cannot change it and if there is something factually incorrect they need to raise this with the team. 

**Describe alternatives you've considered**
The team not getting a notification or not allowing late submissions but we get so many late submissions this is not practical for both the assessor/candidate and the team.

crablab:
I'm going to split this into two separate tickets:
1. UI changes in `admin` to call out late IAs
2. A digital platform ticket for a cloud function to send emails when a late IA is submitted

crablab:
- [ ] UI flag to show whether an IA is/was submitted late 
- [ ] UI option to allow/deny the late submission 
- [ ] Record decision in Store 
- [ ] Display decision in UI

PR #727, opened 2020-08-03
#711 - Identifies late IAs in admin

crablab:
Helper functionality:
- New `isDateGreaterThan` date helper with tests
- 'Borrowed' the `assessment` store with modification to allow updating single assessment record 
- Register `assessment` in `store.js` 

Functionality: 
- Running late IAs (ie. those not submitted but past due date) show as `LATE`
- Submitted late IAs (ie. those submitted past due date) show as `LATE`
- Needing approval IAs (ie. those submitted past due date) have download button hidden and an approval button shown/
- Allows approval for late IAs to be approved (which allows them to be downloaded)

Rejection for late IAs is implicit (can be changed - ticket isn't clear on this) 

screenshot_2020-08-03-155509

### DIFF
--- a/src/helpers/date.js
+++ b/src/helpers/date.js
@@ -20,6 +20,10 @@ const isDateInFuture = (date) => {
   return date > today;
 };
 
+const isDateGreaterThan = (dateOne, dateTwo) => {
+  return dateOne > dateTwo;
+};
+
 const formatDate = (date, type) => {
   if (!(date instanceof Date)) {
     throw 'Supplied date must be a Date object';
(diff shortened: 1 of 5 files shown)


crablab:
@ClaireTroughtonJAC Please could you confirm if the `Approve late submission` button should be two separate approve/reject buttons, and if any copy change needed :smile: 

Thanks!

crablab:
As discussed at standup, Claire is happy with this approach and the logic. We've discussed a separate reject button, but we feel rejection is implicit here. We also discussed disabling the approval button after the hard limit, but this isn't ideal if IAs are processed subsequently. 

@warrensearle to review :smile:

Issue #739, opened 2020-08-05
Pensions Information,  Competency Framework, Welsh Translation not properly "plumbed in" in admin


crablab:
The effect of this is: 
- You can upload any of those documents
- You will always download the candidate assessment template

PR #740, opened 2020-08-05
#739 - Fix incorrect idents causing file overwrites

crablab:
This fixes a bug where files were being uploaded with identical idents (ie. file names) and so uploading one of: 
- Pensions Information
- Competency Framework
- Welsh Translation 
would overwrite the Candidate Assessment Form, and all 4 would download the same file.

@warrensearle please could we roll this as a hotfix to admin? We'll also need to communicate this to teams I think @ClaireTroughtonJAC - any ideas?

### DIFF
--- a/src/views/Exercises/Edit/Downloads.vue
+++ b/src/views/Exercises/Edit/Downloads.vue
@@ -60,7 +60,7 @@
 
         <RepeatableFields
           v-model="exercise.downloads.pensionsInformation"
-          ident="candidate-assessement-forms"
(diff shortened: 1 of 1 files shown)


warrensearle:
@crablab yes we can release this as a hot fix. Just to confirm will this break any of the existing uploads... or are they already broken?

crablab:
@warrensearle They aren't "broken", just the 4 listed will have incorrect files attached. Hence we'll need to let people know and get those fixed/reuploaded.

PR #744, opened 2020-08-06
Replace CSV exports with XLSX

YaaL:
* Contacts
* Diversity report
* Reasonable Adjustments report


YaaL:
@warrensearle There's no need to keep the CSV exports, we only had them because we didn't have XSLX.

Issue #761, opened 2020-08-10
QTs: start the initialised QT


crablab:
As an admin I want to be able to open my QT for candidates. 
As an admin I want candidates to know when the QT has been opened. 

--- 

This should be a button with a confirmation (as it's a permanent and high risk action). 

We may want to have a "Start QT" button which takes the user to a confirmation page. 

The key risks are: 
- This emails all the candidates to let them know the QT is open 
- It cannot be undone 

Once confirmed we should update the QT `status` field and trigger a function: https://github.com/jac-uk/digital-platform/issues/345

warrensearle:
Moving to 'blocked' column whilst we decide whether this should be a start button, that needs to be pressed to activate the test, or a pause button that can be pressed to prevent the test from automatically activating

warrensearle:
Have implemented the cloud function activateQualifyingTests. This works out of the box for manually starting the QT and can be adapted the work with a scheduled cloud task instead.

PR #781, opened 2020-08-17
#761 Activate button should call cloud function

warrensearle:
Included a button to call the `activateQualifyingTest` cloud function.

Note a confirmation step/modal is still needed here.


lloback (approved):
I will merge this for you.

Issue #784, opened 2020-08-19
Enable late applications to be processed


warrensearle:
Once admin staff press the 'begin processing applications' button there is no way to include late applications...

Here we will add a basic way for admin staff to process late applications.

Note: we may in the future look at better support for checking and approving whether late applications can be included in exercise processing.

PR #785, opened 2020-08-19
#784 Enable processing late applications

warrensearle:
Includes another (temporary) button!
This one enables admin user to allow late applications to be processed in 'stages & status'.
When we have roles this can be for super admins only. That said there will be a point in the future when existing exercises have all been processed and so this button will no longer be relevant and can be removed.


lloback (approved):
Good to go

tomlovesgithub (approved):
Great solution!

PR #818, opened 2020-08-27
Bugfix/799 tidy up

lloback:
This is a generic fix for the left nav.
We can use this to tidy up the code below on this function `isActive(page){` because this potentially may work for any left-nav item that we want highlighted.


warrensearle (approved):
Naughty but nice 👍

lloback:
> Naughty but nice 👍

I Agree. It is naughty, but it is simpler than `split` then `split` again, then `length`, then `toLowerCase`, etc, etc

Issue #812, opened 2020-08-27
QTs: add solution for SJ and CA test creation


crablab:
As an admin I want to specify the correct solution for my question so responses can be automatically graded. 

--- 

This will make changes to the test builder and review pages on the admin QT creation process. 

For Critical Analysis: 
- A required drop down within the `QualifyingTestQuestion` component allowing selection of one of the supplied options as the correct answer 
- Require at least one option per question 

For Situational Judgement: 
- Two required drop downs allowing selection of one option as "most appropriate" and another as "least appropriate"
- Require at least two options per question 

On the review pages we need to highlight the answer supplied. 

This will be stored in the data model thus: 

```
{
  "type": "situational-judgement",
  "details": "I am the first question"
  "options: [
     0: "I am the first option",
     1: "I am the second option",
     2: "I am the third option",
     3: "I am the fourth option",
   ], 
  "solution": {
    "mostAppropriate": 0, 
    "leastAppropriate": 2,
  }
 }

{
  "type": "critical-analysis",
  "details": "I am the first question"
  "options: [
     0: "I am the first option",
     1: "I am the second option",
     2: "I am the third option",
     3: "I am the fourth option",
   ], 
  "solution": {
    "correct": 0,
  }
 }
```

YaaL:
Can just do `solution: 0` for critical analysis.

crablab:
I wondered, but I thought keeping the structure consistent might be better in case we want to make changes later? (ie. variable value scoring or something)

PR #825, opened 2020-08-31
#812 Add and edit question solutions

warrensearle:
- Tidy up test builder buttons and titles
- New Select component
- QT list should direct to most relevant View page


lloback (approved):
👍

PR #827, opened 2020-09-01
Feature/801 export qt response to xlsx

tomlovesgithub:
Working DL for xlsx for critical analysis and situational judgement responses.
- Nothing for scenarios yet
- liable to change with data model


warrensearle (approved):
Great, lets get this in and tested 👍

PR #834, opened 2020-09-03
Feature/829 create qt before application opened

tomlovesgithub:
- Created exercises have qualifying test option in tasks side nav by default
- Once an exercise has applications IAs and Character checks will appear in same menu

### DIFF
--- a/src/views/Exercises/Show.vue
+++ b/src/views/Exercises/Show.vue
@@ -55,6 +55,11 @@
             title="Tasks"
           />
           <Navigation
+            v-else
+            :pages="exerciseQtNavigation"
+            title="Tasks"
+          />
+          <Navigation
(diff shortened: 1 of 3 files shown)


src/views/Exercises/Show.vue
```diff
@@ -192,21 +197,37 @@ export default {
...
+          },
+          {
+            title: 'Character Checks',
+            name: 'exercise-tasks-character-checks',
+          },
+        );
+      }
+      return tasks;
+    },
+    exerciseQtNavigation(){
```
warrensearle, line 223: Think you can remove this now

tomlovesgithub: 🤦‍♂️

lloback (approved):

Issue #860, opened 2020-09-16
Date Input component not storing datetime correctly


warrensearle:
The DateInput component is not working when in datetime mode (see screen grab).

Enter a date or time and it straight away replaces what was entered with 01:00am on 25 October!

![Screenshot 2020-09-16 at 22.37.37.png](https://images.zenhubusercontent.com/5e412648a39708a606c962c9/cc4d3c9b-2bc7-48cf-8aba-57f2e8522c89)

warrensearle:
See related bug #848

warrensearle:
Confirmed as on staging and ready for review

Issue #848, opened 2020-09-14
Website Listing:  Launch dates defaulting to 29 March 2020


ClaireTroughtonJAC:
**Describe the bug**
On admin the website listing function is not accepting any date entered.  Whatever date is added to the "launch fields" is defaulting to 29 March 2020 when you save.

**To Reproduce**
1. Go to 'JAC00020
2. Click on 'Website listing'
3. Scroll down to 'launch date' 
4. Enter 01/10/20
5. Press save and continue 

Date returns to 29/03/2020

**Expected behavior**
The actual launch date should be displayed here and in the vacancies list on Apply

**Screenshots**
![Screenshot 2020-09-14 at 16.00.03.png](https://images.zenhubusercontent.com/5e3c2b35e9bf7c0ea1c2859a/19fa7872-a2c2-408f-90de-b16143a2391a)

Issue #831, opened 2020-09-02
QT start and end time are not displaying as entered


warrensearle:
Date and times as entered for QT:
![Screenshot 2020-09-02 at 12.20.17.png](https://images.zenhubusercontent.com/5e412648a39708a606c962c9/b10b5f55-d74d-4f07-a66a-37bfb5538578)

Date and times as displayed on Review page:
![Screenshot 2020-09-02 at 12.21.08.png](https://images.zenhubusercontent.com/5e412648a39708a606c962c9/7f2cd035-bd2e-4ac1-abba-0bfe737b45a1)

We need to 
- [x] Check that times are stored in firestore correctly
- [x] Ensure times are displayed correctly on Review page

PR #861, opened 2020-09-16
#860 Fix DateInput component not allowing datetime

warrensearle:
This removes again the `Date.UTC` call that was added in #848 and removed in #831.

Instead I have changed the default time to 13:00. Partly because JAC tends to use 13:00 as default time for launching an exercise, for example. Mainly because this means we bypass the BST daylight saving issue that was being experienced where a default time of 0:00 was (when in BST) being converted to 23:00 the previous day.

### DIFF
--- a/src/components/Form/DateInput.vue
+++ b/src/components/Form/DateInput.vue
@@ -206,9 +206,8 @@ export default {
       const day = this.type === 'month' ? 1 : this.day;
       const month = this.month;
       const year = this.year;
-      const hour = this.type === 'datetime' ? this.hour : 0;
+      const hour = this.type === 'datetime' ? this.hour : 13; // default time to 1pm. this avoids BST issue we would have if we defaulted to 0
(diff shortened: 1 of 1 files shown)


warrensearle:
@lloback and @tomlovesgithub I have merged this into develop as it's needed for tomorrow's Qualifying Test dry run however please do both review this change.

src/components/Form/DateInput.vue
```diff
@@ -219,7 +218,7 @@ export default {
         if (this.dateConstructor === null) {
           return null;
         } else {
-          return new Date(Date.UTC(...this.dateConstructor));
+          return new Date(...this.dateConstructor);
```
lloback, line 221: I have changed this to UTC because we were having issues with 29/03/2020 which is the date that GMT becomes BST. WE need to test this.

lloback: @warrensearle

PR #866, opened 2020-09-17
Feature/864/qt survey field admin

tomlovesgithub:
Add qualifyingTest.feedbackSurvey for feedback form URL
A few Q's on how this is going to be available in the applicant view, and if the url will be copied to the response object or we will have to access the qualifyingTest object directly


src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Edit.vue
```diff
@@ -56,6 +56,12 @@
         required
       />
 
+      <TextField
+        id="qualifyingTest-testDuration"
```
lloback, line 60: @tomlovesgithub you need to change this ID to `qualifyingTest-feedback-survey`.
You can't repeat IDs on the page.

lloback (changes requested):

warrensearle (changes requested):
Please can you add the new field `feedbackSurvey` to the defaults object, with a null value (within the data function)

lloback:
@tomlovesgithub You have 40 min to get this into the daily release ... no pressure.

lloback (approved):
ready to go.

lloback:
> Please can you add the new field `feedbackSurvey` to the defaults object, with a null value (within the data function)

this was done.

Issue #662, opened 2020-07-13
Ability for Admins to upload candidate assessments to an application


ClaireTroughtonJAC:
**Is your feature request related to a problem? Please describe.**
We often receive candidate assessment via email when candidates experience issues uploading their assessment or statement of suitability. 

**Describe the solution you'd like**
I want the admins to have the ability to edit a candidate's application - prior to close or after it - to upload this document and display it in their Full application.  

**Describe alternatives you've considered**
The Digital team doing it

**Additional context**
This will prevent issues of teams needing to ask the Digital team to add the assessments directly

Essentially add a  "change" button followed by an "upload" button so that the admin can attach the correct file.
<img width="1039" alt="Screenshot 2020-07-13 at 12 44 27" src="https://user-images.githubusercontent.com/27224574/87300934-a1029c80-c506-11ea-89c3-20d83c3aa6ec.png">

The following were moved to ticket #871

**Technical notes**
Once we allow multiple users to attach files to an application, we need to track who and when.
This metadata doesn't belong to an application so should be kept in a separate `files` collection.
The `file` document should contain at least:
* upload timestamp
* id of owner (user who uploaded it)
* file reference (path to firestore or s3 if we decide to switch)
* original filename 
* filename for download
* document type (`cv`, `self-assessment`, `independent-assessment`, etc.)
* eventually virus check status + timestamp
* optionally file size

Applications, exercises and other documents should then reference this entry instead of storing file reference.

ClaireTroughtonJAC:
@YaaL I have added the part about keeping the history of files.  You wanted to add your idea about Metadata

warrensearle:
Hey @lloback we discussed splitting this into two issues:

One to fulfill the immediate needs of allowing admin staff to upload assessments on behalf of Assessors (e.g. they've been emailed a file and need to add it on their behalf). Suggest this ticket is used for that piece of work.

The other ticket is to capture the great plan outlined in the Technical Notes above: to track multiple file versions, who, what, where, when etc. When this is finalised it can then be rolled out for all uploads.

lloback:
Allow the admins to upload a file on behalf of the candidate. Split the ticket
1. simple upload version
2. "The other ticket is to ..."

lloback:
Split the ticket into 2: #871

PR #867, opened 2020-09-17
Feature/662 upload

lloback:
#662 Let admins upload Self Assessment Competencies AND Statement of Suitability on behalf of the candidates.
This is half of the task.


warrensearle (approved):
👍

PR #869, opened 2020-09-18
Provide custom PJE days

tomlovesgithub:
image
Make admins provide amount of days experience when adding/editing eligibility info
image
And show this on the show eligibility page


warrensearle (approved):
👍

PR #889, opened 2020-09-22
884/view scenario test

tomlovesgithub:
<h1> cant connect to issues cause i dont have zenhub chrome plugin on this laptop - sorry!! </h1>

Theses views have been built assuming the following JSON structure which is based on [this firestore entry](https://console.firebase.google.com/u/1/project/digital-platform-develop/firestore/data~2FqualifyingTestResponses~2FNZlaCSysvy55Thw7FUAx) 
```
...
testQuestions: {
  introduction: "I am introductory instruction text",
    questions: [
       0: {
        details: "I am the scenario to work through",
        questions: [
                   0: {answer: "Question 1", limit: 100},
                   ...
       },
          ...
     ],
        responses: [
                   0: {text: "I am the text the candidate entered", count: 50, started: datetime, completed: datetime}
                   1: {text: null, count: 0, started: datetime, completed: datetime} // Null indicates skipped question
        ]
  }
 }
...
```
I have some concerns over where `this.qualifyingTest.testQuestions.questions[i].questions[i].options[i].text` is used as i wasn't sure weather this value should rather come from the response object. 
I have come to the conclusion that it should be okay as its on the responses page, where the XLSX is built so has access to both objects, just threw me a little.
All the same i am a unsure of weather i have interpreted the data objects incorrectly with the header titling on responses page line 97 (will highlight on PR)

Some of the response objects don't have the `element.application.referenceNumber` I'm assuming this is a development issue rather than a legacy thing, so my solution doesn't care that this doesn't exist, may error on staging/dev if not downloading the correctly formatted data.

Im using v-html, i got some warnings and noticed that eslint has been disabled else where it is used.


src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
```diff
@@ -92,9 +92,11 @@ export default {
         if (this.qualifyingTest.type === QUALIFYING_TEST.TYPE.SITUATIONAL_JUDGEMENT) {
           headers.push(`Q ${ index + 1 }. Most Appropriate`, `Q ${ index + 1 }. Least Appropriate`);
         }
-        // if (this.qualifyingTest.type === QUALIFYING_TEST.TYPE.SCENARIO) {
-        //   headers.push('scenario');
-        // }
+        if (this.qualifyingTest.type === QUALIFYING_TEST.TYPE.SCENARIO) {
+          question.options.forEach((option, decimal) => {
```
tomlovesgithub, line 96: unsure about these :?

lloback: No worries, I fixed it.
- decimal is the `index`

lloback (approved):

PR #907, opened 2020-09-28
Fix reference number in qt score download

tomlovesgithub:
Ref num for QT downloads is sometimes missing and when working on a feature i hardcoded it to a string, ive now added a fallback default empty string which will show when the field isn't present and the same thing for the fullName field which will default to the name field


warrensearle (approved):
Looks good to me 👍

lloback (approved):

Issue #820, opened 2020-08-27
QT: DELETE a Qualifying test


lloback:
AS an admin I want to be able to DELETE a Qualifying test if it is on DRAFT mode or not started.

- What status do I want the QT to be so I can delete?
- Check that it deletes all the questions and responses, etc?

PR #909, opened 2020-09-28
#820 Delete a QT

warrensearle:


src/store/auth.js
```diff
@@ -30,11 +30,23 @@ const module = {
...
+          if (
+            [ // TODO User roles!
+              'warren.searle@judicialappointments.digital',
+              'tom.russell@judicialappointments.digital',
+              'lisias.loback@judicialappointments.digital',
+              'rebecca.mcknight@judicialappointments.digital',
+            ].indexOf(user.email) >= 0
+          ) {
+            role = 'superadmin';
+          }
```
lloback, line 43: Not great to have personal details on the code ... CAn this be done using the Google Auth?

warrensearle: Yes, this is a very temporary fix so that we four can remove old tests.
Starting tomorrow we'll be working through user roles and auth tokens properly... so expect this will be fixed properly in the next few days

lloback: good.

lloback (approved):

PR #910, opened 2020-09-29
Feature/903 data model

lloback:
Change the View for 
- Scenario Test
- Situational Judgement
-  Critical Analysis


warrensearle (approved):
👍

PR #923, opened 2020-10-02
901/reasonable adjustment justification textarea

tomlovesgithub:
Updated the editable field component to include textarea type, and added a conditional to only show reasonable adjustments justification if an adjustment has been made


warrensearle (approved):
Does the trick 👍

Issue #838, opened 2020-09-07
Indicate whether a candidate went offline during the test 


ClaireTroughtonJAC:
See ticket apply#481 for how we log a user's presence during a qualifying test.

---

As an admin
I would like to see evidence of whether a candidate went offline during a test
so that I can make an informed decision about whether to allow the candidate to re-sit or to give them special consideration.

---

On the qualifying test response page we should include a new tab called 'Connection Log', say, which will display the user's connection log during the test.

Data is available from Realtime Database at the path `qualifyingTest/{qualifyingTestId}/userStatus/{userId}`
Each document contains an `online` timestamp indicating when the user connected to the database and an `offline` timestamp indicating when the user went offline.
Combining data from all the documents we can present a picture of how long the user was online vs offline, when they were last seen etc.

Note that a user will be marked as offline under the following conditions:
- losing connection
- closing browser or browser tab
- refreshing the browser page
- completing the test

The user will not be marked as offline when they follow the 'exit test' link

lloback:
Added 2 things to this task:

- [ ] Added the OFFLINE time in between
- [ ] Added information about the user being ONLINE just now

PR #926, opened 2020-10-05
#838 Time Logs

lloback:
- Added the OFFLINE time in between
- Added information about the user being ONLINE just now


warrensearle (approved):
Looks good to me 👍

PR #928, opened 2020-10-06
Remove duplicate answers response review page

tomlovesgithub:
Tried setting up a test suite for this view but encountered a-lot of issues removing warnings from the suite cause i couldnt satisfactorily mock the full qt object


warrensearle (approved):
👍

PR #944, opened 2020-10-09
Editable field email fix

tomlovesgithub:
This came up in technical troubleshooting 9/10/20, and was an issue i think i introduced when making changes to the editable field component to have use textareas


warrensearle:
Changes have been moved to hotfix/0.36.2

PR #952, opened 2020-10-13
Changed data model for invitations

mbrookeswebdev:


src/store/invitations.js
```diff
@@ -33,25 +33,22 @@ export default {
...
               candidate: {
                 email: email,
                 id: null,  // populated when candidate updates the invite
               },
               status: 'created',  // 'created' | 'invited' | 'accepted' | 'rejected'
               statusLog: {
-                created: firebase.firestore.FieldValue.serverTimestamp(),
-                invited: null,  // populated when email invite has been sent (out of scope right now)
+                created: serverTimestamp,
+                invited: serverTimestamp,
```
warrensearle, line 51: `invited` should remain as `null`. It will be set if/when we include sending out email invitations

warrensearle (changes requested):
Just one change please

src/store/invitations.js
mbrookeswebdev: Tha't done.

lloback (approved):

warrensearle (approved):

Issue #873, opened 2020-09-21
Stage/status selection when initialising a qualifying test


warrensearle:
Generally tests will be initialised for all candidates in the Review stage (less any candidates that have withdrawn)

It should however also be possible to refine the selection down to only those candidates with a particular status.

For example, following the results of 'Test A' the applications will be marked with a status of 'Passed Test A' or 'Failed Test A'. For 'Test B' it should be possible to initialise tests for only those applications which 'Passed Test A'.

Note in ticket [TBC] we will be allowing test results to be combined and then marking applications with a status for the combined results.
For example, 'Test A' scores will be added to 'Test B' scores. Those applications with satisfactory total score will be marked with a status of, say, 'Passed Test A + Test B'.
A subsequent test will be able to be initialised from those applicants with status of 'Passed Test A + Test B'.

PR #953, opened 2020-10-13
#873 Stage/status selection when initialising a qualifying test

lloback:
image


warrensearle (approved):
Let's get this into develop env. and tested 👍

Issue #932, opened 2020-10-06
Download qualifying test report


warrensearle:
Enable qualifying test report data to be downloaded as an excel spreadsheet.

Also ensure on-screen report enables drill-down into each score

PR #954, opened 2020-10-13
#932 Download qualifying test report

warrensearle:
Also includes drilling down into each score


lloback (changes requested):
[Vue warn]: Error in render: "TypeError: Cannot read property 'female' of undefined"

lloback:
I don't kno if this is expected, but the percentage is NaN
image

warrensearle:
@lloback I've removed old data from our develop database, which was responsible for the issues experienced above.

lloback (approved):
all good.

PR #964, opened 2020-10-20
Feature/962 withdrawn

lloback:
Withdrawn applications should be removed from IAs and QTs

### DIFF
--- a/src/store/application.js
+++ b/src/store/application.js
@@ -2,6 +2,7 @@ import firebase from '@firebase/app';
 import { firestore } from '@/firebase';
 import { firestoreAction } from 'vuexfire';
 import vuexfireSerialize from '@/helpers/vuexfireSerialize';
+import { STATUS, QUALIFYING_TEST_RESPONSE } from '@/helpers/constants';
 
 const collection = firestore.collection('applications');
 
@@ -71,6 +72,17 @@ export default {
           });
         }
       }
+    },
+    withdrawn: async (context, data ) => {
+      const applicationId = data.applicationId;
+
+      await context.dispatch('update', { data: { status: STATUS.WITHDRAWN }, id: applicationId });
+
+      //  If IAs has started ensure relevant assessments documents are removed (soft deleted)
+      context.dispatch('assessment/delete', { data: { status: STATUS.DELETED }, id: applicationId }, { root: true });
+
+      // If Qualifying Tests have started ensure the relevant qualifyingTestResponse document is removed (soft deleted)
+      context.dispatch('qualifyingTestResponses/delete', { data: { status: QUALIFYING_TEST_RESPONSE.STATUS.DELETED }, id: applicationId }, { root: true });
     },
   },
   state: {
--- a/src/store/application.js
+++ b/src/store/application.js
@@ -2,7 +2,7 @@ import firebase from '@firebase/app';
 import { firestore } from '@/firebase';
 import { firestoreAction } from 'vuexfire';
 import vuexfireSerialize from '@/helpers/vuexfireSerialize';
-import { STATUS, QUALIFYING_TEST_RESPONSE } from '@/helpers/constants';
+import { STATUS } from '@/helpers/constants';
 
 const collection = firestore.collection('applications');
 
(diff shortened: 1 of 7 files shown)


src/store/application.js
```diff
@@ -72,6 +73,17 @@ export default {
         }
       }
     },
+    withdrawn: async (context, data ) => {
+      const applicationId = data.applicationId;
+
+      await context.dispatch('update', { data: { status: STATUS.WITHDRAWN }, id: applicationId });
+
+      //  If IAs has started ensure relevant assessments documents are removed (soft deleted)
+      context.dispatch('assessment/delete', { data: { status: STATUS.DELETED }, id: applicationId }, { root: true });
```
warrensearle, line 82: I think the call to 'assessment/delete' should not include `data`. It should just be the ID to delete. The delete method in that store should take care of implementing what 'delete' means by updating `status` and `statusLog` itself (if that's what it does)

src/store/application.js
```diff
@@ -72,6 +73,17 @@ export default {
...
+
+      // If Qualifying Tests have started ensure the relevant qualifyingTestResponse document is removed (soft deleted)
+      context.dispatch('qualifyingTestResponses/delete', { data: { status: QUALIFYING_TEST_RESPONSE.STATUS.DELETED }, id: applicationId }, { root: true });
```
warrensearle, line 85: Same as comment about 'assessment/delete'

src/store/qualifyingTest/qualifyingTestResponses.js
```diff
@@ -67,6 +67,26 @@ export default {
...
+      const collection = firestore.collection('qualifyingTestResponses')
+        .where('application.id', '==', id)
+        .get()
+        .then(async snapshot => {
+          // console.log('snapshot', snapshot);
+          // eslint-disable-next-line no-unused-vars
+          snapshot.forEach(response => {
+            // console.log('response.id', response.id);
+            const ref = firestore.collection('qualifyingTestResponses').doc(response.id);
+            batch.update(ref, data);
```
warrensearle, line 84: I don't understand this bit. Why is there an update followed by another one (using set)?

src/store/stage/review.js
```diff
@@ -108,6 +108,13 @@ export default {
         batch.update(ref, data);
       });
       await batch.commit();
+
+      if (status === APPLICATION_STATUS.WITHDREW_APPLICATION) {
+        selectedItems.map( async item => {
+          // call withdraw applicationstore 
+          await context.dispatch('application/withdrawn', { applicationId: item }, { root: true });
```
warrensearle, line 115: Being pedantic now but feel this action should be a verb or a more descriptive name.
e.g. `withdraw` rather than `withdrawn`.... or perhaps `withdrawApplication`

warrensearle (changes requested):
Few tweaks. See comments

src/store/application.js
lloback: done.

src/store/application.js
lloback: done.

src/store/qualifyingTest/qualifyingTestResponses.js
lloback: Basically, the UPDATE and the SET have different behaviours.
UPDATE, replaces all the object in the database.FOr instance, if I try to update

```
        statusLog: { 
          'deleted': timestamp,
        },
```

All the other children nodes of `statusLog` will be overwritten, so, for example the following would disappear and be replaced by the previous code

```
        statusLog: { 
          'created': timestamp,
          'started' .......
        },
```

On the other hand, SET with the merge flag will keep it.

lloback: So, removed the UPDATE and used SET instead.

src/store/stage/review.js
lloback: you know it looks the same to me - so, thanks for pointing out!
Changed.

lloback:
@warrensearle this is ready for review.

src/store/qualifyingTest/qualifyingTestResponses.js
warrensearle: Ah right. Did you know you can use dot paths like:
```
data = {
  'statusLog.deleted': timestamp,
};
```
This will leave the rest of `statusLog` object in tact

warrensearle (approved):
Noice 👍

Issue #963, opened 2020-10-20
Add optional text area to questions for Scenario test for instructions 


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
As an admin I want to be able to add question instructions that expand upon the question. 

**Describe the solution you'd like**
A title box for the question. An optional text area for instructions under this that maintains any line breaks the admin insert. This will allow the team for Recorder to put question 2 in and then put the instructions saying for this question please.... with further hints to the candidate. 

This text box needs to be
1) optional
2) allow admin to use line breaks
3) display line breaks on apply 
4) be an option for each Q 

Need before 4:00pm on 21st Oct.

PR #966, opened 2020-10-21
#963 Add optional text area to questions for Scenario test for instru…

lloback:
image


warrensearle (approved):
Looks good 👍 
(only question is whether hint is the right name for this)

lloback:
It needs to be, because it is replacing the HINT line "Answer below:"

warrensearle:
Is it? I thought it was optional and in addition to the existing 'Answer below:' hint. Please check with @Rebecca-mcknight1991

lloback:
> Is it? I thought it was optional and in addition to the existing 'Answer below:' hint. Please check with @Rebecca-mcknight1991

I thought the same, but then, when I looked into the code, There was no place to add it as it was heavily embedded on the  TEXTINPUTAREA ...

Issue #967, opened 2020-10-21
QT results - see those who did not submit 


Rebecca-mcknight1991:
**Is your feature request related to a problem? Please describe.**
As an admin I need to be able to distinguish between the candidates who never started a QT and those who did. 

**Describe the solution you'd like**
Add a status column which includes: Started, Completed, Completed but out of time and Most importantly 'Activated' as these are the ones who never started the QT and therefore did not submit. They get a separate email. 

**Additional context**
Think they might need this quite soon but I'll let you know Warren. 

---

This has a corresponding task TBC

PR #971, opened 2020-10-22
#967 QT results - see those who did not submit

warrensearle:
Includes status in the QT report download

Also auto-fixed some rogue spaces


lloback (approved):
approved

Issue #1018, opened 2020-10-30
Draft, applied and withdrawn links not showing in side navigation


Rebecca-mcknight1991:
When I click begin processing apps on staging the applications section then does not appear when I know there are draft and sent applications.

warrensearle:
We have re-factored the logic which determines when the Applications side navigation displays... So hopefully this bug has been resolved as a result.

PR #1019, opened 2020-10-30
#1018 Draft, applied and withdrawn links missing from side navigation

warrensearle:


lloback (approved):

tomlovesgithub (commented):
Looks good to me, would be great to turn the tests on for the views but more work needs to be done adding to the unit test helper

lloback (approved):

PR #1027, opened 2020-11-02
#74 Prevent changes to IA dates after assessments were created

mbrookeswebdev:


src/components/Form/DateInput.vue
```diff
@@ -143,6 +146,10 @@ export default {
   },
   extends: FormField,
   props: {
+    assessmentsInitialised: {
```
warrensearle, line 149: I think you could call this prop `disabled` which will then mean that the `DateInput` component implements the html `disabled` attribute. Which would be nice and something we might roll out to other form components 👍

warrensearle (changes requested):
This is great. Just see the suggested prop name change 👍

src/components/Form/DateInput.vue
mbrookeswebdev: Done! 👍

warrensearle (approved):
Lovely 👍

Issue #594, opened 2020-06-16
No existance checking on exercise.estimatedLaunchDate


crablab:
**Describe the bug**
`formatEstimatedDate` is being called in `/exercises/show` with an undefined value (`estimatedLaunchDate`). Assumption is this is because `estimatedLaunchDate` can be undefined in some cases. 

**To Reproduce**
Open an exercise which does not have an estimated launch date. 

**Expected behaviour**
If `exercise.estimatedLaunchDate` is undefined, a placeholder should be shown instead.  

[Sentry](https://sentry.io/organizations/judicial-appointments-commissi/issues/1611168302/events/latest/?project=1792541&query=is%3Aunresolved)

PR #1060, opened 2020-11-10
#594/Scenario moveTest tests and mocks

tomlovesgithub:
Its going to worth adding to the test suite created here for the other QT types and spending some time documenting the how we expect this stuff to work, so that tests can be added to.


warrensearle (commented):
This is looking good. I think `moveTest` does the same for all test types so we shouldn't need to repeat the tests for each type

src/store/qualifyingTest/qualifyingTestResponses.js
```diff
@@ -99,33 +99,16 @@ export default {
...
-                completed: null,
-              };
-            }
-            responses.push(response);
-          });
-          data.responses = responses;
-        }
-      }
+
       await context.dispatch('update', { data: data, id: qualifyingTestResponse.id });
```
lloback, line 112: @mbrookeswebdev are you sure that removing this will continue to work backwardly-compatible? What steps did you take to test this?

lloback (changes requested):

src/store/qualifyingTest/qualifyingTestResponses.js
mbrookeswebdev: @lloback I think you meant to tag @tomlovesgithub

tomlovesgithub: https://app.zenhub.com/workspaces/platform-development-5ea838cd2aec471eb6d14139/issues/jac-uk/apply/594#issuecomment-724053152
tests have been added, removing the block was part of the requirements

lloback (approved):

warrensearle (approved):
👍

Issue #1055, opened 2020-11-09
Held fee-paid judicial role needs to be a column heading in export from applied


Rebecca-mcknight1991:
As an admin I need to be able to clearly see who has and has not held a Held fee-paid judicial role. The only place this is captured is in the equality and diversity questions where it asks Have you held fee-paid judicial role? The export from applied needs to have this information as one of the headings and capture it as yes/no 

https://admin.judicialappointments.digital/exercises/D7vODO7AVyTTlctVmA5Y/applications/applied
Export contacts

Rebecca-mcknight1991:
@warrensearle could you get someone on this fix - please let me or @joy-ade know if any questions.

mbrookeswebdev:
Confirmed with @Rebecca-mcknight1991 , need to include Prefer not to say option as well.

mbrookeswebdev:
Confirmed with @Rebecca-mcknight1991 , use 'Unknown' for empty/null or undefined values.

PR #1061, opened 2020-11-10
#1055 Added Held fee paid judicial role into the export file

mbrookeswebdev:


tests/unit/filters.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  it('returns string `Prefer not to say`', () => {
+    expect(heldFeePaidJudicialRole('prefer-not-to-say')).toBe('Prefer not to say');
+  });
+});
+
+describe('when input value is boolean `false`', () => {
+  it('returns string `No`', () => {
+    expect(heldFeePaidJudicialRole(false)).toBe('No');
+  });
+});
```
warrensearle, line 31: What is expected output when input value is empty/null?

warrensearle (approved):
This looks good. 👍  
See comment/question with regards empty/null value.

tests/unit/filters.spec.js
mbrookeswebdev: Yes, I was not sure how to deal with that. Technically we could only receive five pre-defined values - (three types of judicial role (as a Yes), No and prefer-not-to-say). At the moment empty/null is covered by 'Prefer not to say', which is essentially 'Unknown'. Should I cover this case in the function and return 'Unknown' if somehow value is null/empty?

warrensearle: I'm thinking the filter should probably return an empty string when input is null/empty and only return 'Prefer not to say' when the candidate has chosen that option. Please double check with Becky if you like

mbrookeswebdev: @warrensearle I've checked with Becky, she said to return 'Unknown', so I've made the changes.

warrensearle (approved):
Excellent, thanks @mbrookeswebdev 👍

PR #1062, opened 2020-11-11
Feature/960 upload Assessment on Behalf of the assessor

lloback:
This functionality allows the admns to upload a file on behalf of the assessor.


mbrookeswebdev (approved):

tomlovesgithub (approved):

warrensearle (changes requested):
This is looking good however I think we need to improve the data model. Previously only Assessors uploaded assessments and we used their `AssessorId` in the storage path for where the file is saved. Now that admin can upload an assessment on behalf of the assessor we don't (necessarily) have the `AssessorId` and we shouldn't populate this with the Admin user's UID instead - unless we are clear that's what we've done and even then it feels wrong.

Perhaps we should store the explicit storage path in the assessment document itself.

lloback:
> This is looking good however I think we need to improve the data model. Previously only Assessors uploaded assessments and we used their `AssessorId` in the storage path for where the file is saved. Now that admin can upload an assessment on behalf of the assessor we don't (necessarily) have the `AssessorId` and we shouldn't populate this with the Admin user's UID instead - unless we are clear that's what we've done and even then it feels wrong.
> 
> Perhaps we should store the explicit storage path in the assessment document itself.

This is done.

warrensearle (approved):
This works well 👍

PR #1069, opened 2020-11-17
#985 Independent assessor change

lloback:
Change assessor

Changes the `application` and `assessor` if exists

### DIFF
--- a/src/views/Exercises/Applications/Application.vue
+++ b/src/views/Exercises/Applications/Application.vue
@@ -1547,19 +1547,26 @@
             <div class="govuk-!-margin-top-9">
               <h2 class="govuk-heading-l">
                 Independent assessors
-              </h2>
+              </h2> 
 
               <dl class="govuk-summary-list">
+                <div class="govuk-summary-list__row text-right print-none button-right">
+                  <dt class="govuk-summary-list__key" />
+                  <dd class="govuk-summary-list__value">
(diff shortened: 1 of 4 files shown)


src/views/Exercises/Applications/Application.vue
```diff
@@ -2175,6 +2177,34 @@ export default {
         this.$store.dispatch('application/update', { data: { [field]: val }, id: this.applicationId });
       }
     },
+    editAssessor(AssessorNr) {
+      if (AssessorNr === 1) {
+        this.assessorDetails = {
+          AssessorNr: AssessorNr,
+          applicationId: this.application.id,
+          firstAssessorEmail: this.application.firstAssessorEmail,
```
warrensearle, line 2185: I think you could pass more generic data/props to the modal like `email`, `name` etc along with `assessorNumber` which should mean the modal form and the store can be simplified.

warrensearle (commented):
Please see comment. Whilst this is here in a PR I think we should simplify

src/views/Exercises/Applications/Application.vue
lloback: This has been implemented - a cleaner form

warrensearle (approved):
Looks good to me 👍

mbrookeswebdev (approved):

Issue #970, opened 2020-10-22
Update judicial salary bands 


Rebecca-mcknight1991:
The judicial salary bands which you choose in Vacancy information under Appointment type have been changed. They could change quite a bit so we want to change the way they are to look like the Fee-paid section. When you click fee-paid a box appears for you to write the digits of the fee in with a pound sign before it. We need to same for Salaried roles except when you click salaried and the box appears for the figure it says 'Salary'. Do not worry about this updating everywhere. Just before we push onto staging I need to get Matt Howley to tell everyone on the teams they will need to go into exercises and update the salary. 

**Current way it looks**
![Screenshot 2020-10-22 at 15.46.29.png](https://images.zenhubusercontent.com/5e4bbb99c920b007684ed9ba/d53f5b71-db7b-49c2-a2be-7ba7f6ecd647)

**How it should look - like fee-paid**
![Screenshot 2020-11-17 at 08.25.35.png](https://images.zenhubusercontent.com/5e4bbb99c920b007684ed9ba/8aeec36c-eca5-48ba-906d-06473f072489)

warrensearle:
This Blocked while it is decided whether to update the current set of salary groups/bands or to use a simple free-entry field instead, meaning Admin can enter any salary.

warrensearle:
This is not blocked anymore

PR #1074, opened 2020-11-19
#970 Update judicial salary bands

lloback:
- New salary as text
image

- All new exercises will show a free text field where they can out the salary (no bands anymore)
image

# THIS IS BACKWARD COMPATIBLE
Old exercises will show the Salary bands

image
image

### DIFF
--- a/src/components/Form/Currency.vue
+++ b/src/components/Form/Currency.vue
@@ -23,7 +23,7 @@
     <input
       :id="id"
       v-model="currencyInput"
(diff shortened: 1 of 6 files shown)


warrensearle (approved):
Awesome 👍 
but what's up with the Currency Input? Can it be made wider in this instance?
Screenshot 2020-11-20 at 09 22 38

lloback:
> Awesome 👍
> but what's up with the Currency Input? Can it be made wider in this instance?

I increased the width. Thanks.

PR #1080, opened 2020-11-24
1079/created-at-fix

tomlovesgithub:
BEFORE: 
image

AFTER:
image


src/views/Exercises/Applications/Application.vue
```diff
@@ -98,7 +98,7 @@
             <div class="panel govuk-!-margin-bottom-9 govuk-!-padding-4 background-light-grey">
               <span class="govuk-caption-m">Created on</span>
               <h2 class="govuk-heading-m govuk-!-margin-bottom-0">
-                {{ application.createdAt | formatDate | showAlternative("Unknown") }}
+                {{ new Date(application.createdAt) | formatDate | showAlternative("Unknown") }}
```
warrensearle, line 101: Feels like converting the value to a date should happen within the `formatDate` filter.
We can then ensure we have tests for all possible inputs and get the expected output

warrensearle (changes requested):
Feels like converting the value to a date should happen within the `formatDate` filter.
We can then ensure we have tests for all possible inputs and expected outputs

tests/unit/filters/formatDate.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    const invalidTypes = [
+      ['string'],
+      [null],
+      [undefined],
+      [[]],
+    ];
+    
+    it.each(invalidTypes)('when given %o', (value) => {
+      expect(formatDate(value)).toBe('Invalid Date');
+    });
```
warrensearle, line 41: Are we sure we want `null` or `undefined` to be converted into `'Invalid Date'` string?

warrensearle (commented):
Wow, lots of tests added.

warrensearle (approved):
Let get this made 👍

Issue #1093, opened 2020-11-27
CircleCi needs access to Github Package Registry


lloback:
Added JAC-kit to the the admin. Prbably we need to update the  Github token.

https://app.circleci.com/pipelines/github/jac-uk/admin/2449/workflows/45cee50f-387f-4fff-8a47-c688007db74d/jobs/6088

```
npm ci
npm ERR! code E401
npm ERR! Unable to authenticate, need: Basic realm="GitHub Package Registry"

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2020-11-27T01_01_37_656Z-debug.log


Exited with code exit status 1
CircleCI received exit code 1
```

warrensearle:
This is tested working on staging

PR #1095, opened 2020-11-27
#1093 Update CircleCI config to access Github Packages registry

warrensearle:
This is a rather simple fix, in the end.

1. Create a Personal Access Token (PAT) in Github solely with 'read:packages' access
Screenshot 2020-11-27 at 12 33 31

2. Add a new environment variable in CircleCI called `GITHUB_PACKAGES_TOKEN` and copy the PAT into it
3. Update CircleCI `config.yml` to append the token to our `.npmrc` config file at run-time
4. Continue with install, test, build as before

Note: unit tests need fixing but that's outside the scope of this ticket :)


mbrookeswebdev (approved):

tomlovesgithub (approved):
Awesome, nice that the problem is what we thought

lloback (approved):

Issue #1104, opened 2020-12-07
Move components and shared code to jac-kit


warrensearle:
This is a sister ticket to jac-kit#4 to ensure we remove shared code and update local code to use jac-kit instead

warrensearle:
This is a sister ticket to jac-kit#4 to ensure we remove shared code and update local code to use jac-kit instead

warrensearle:
This is a large but non-functional change: we should test that everything works the same as before.

PR #1105, opened 2020-12-07
#1104 Move components and shared code to jac-kit

warrensearle:
Lots of changes here!

@tomlovesgithub please can you check this out and give it a thorough review.


tomlovesgithub (approved):
lot of changes as you say but looks fine locally here, made sure to test specific things in places where they are coming from the kit too
think its safe to merge 👍

src/filters.js
```diff
@@ -1,4 +1,4 @@
-import { APPLICATION_STATUS, QUALIFYING_TEST } from '@/helpers/constants';
+import { APPLICATION_STATUS, QUALIFYING_TEST } from '@jac-uk/jac-kit/helpers/constants';
```
lloback, line 2: didn't we move the filters???
We can separate this file into some reusable and some that are not and then import it here from the jac-kit.

And then, we can do the same on the apply side - import only the ones that are reusable.

lloback (approved):
@warrensearle I commented on some of the files, but it is good to go. Just read them as a side note...

src/filters.js
warrensearle: Agreed, am in the process of moving the filters and related tests

Issue #996, opened 2020-10-29
Have a search across all lists/tables/reports


Rebecca-mcknight1991:
As an admin I want to be able to search from multiple places. I want to be able to search across the platform which can be done currently if I click into 'candidates'. This should remain. But the search also needs to be in each list and report across admin.

PR #1109, opened 2020-12-15
#1001 Add filtering to table component

warrensearle:
Includes latest Table component from jac-kit and enables filtering for Dashboard and Review list.

Note: remaining Filters to be added in #996


tomlovesgithub (approved):
Nice, seeing the benefits of the JAC-kit!

PR #1138, opened 2021-01-11
Epic/1112 changes to character information

mbrookeswebdev:


warrensearle (approved):
Looks good, lets get this merged and tested on staging.

Many of the components look very similar to those on Apply. Please could you write a ticket for us to look at merging the components so we only have one set rather than two very similar sets.

PR #1190, opened 2021-02-24
account for missing info in applications

tomlovesgithub:
Fixed the issue sentry highlighted and made things look less like:
image

and more like:
image

### DIFF
--- a/src/views/Exercises/Applications/Application.vue
+++ b/src/views/Exercises/Applications/Application.vue
@@ -765,7 +765,7 @@
                   <dd
                     class="govuk-summary-list__value"
                   >
-                    {{ application.applyingForWelshPost | toYesNo | showAlternative('Answer not supplied') }}
+                    {{ application.applyingForWelshPost | toYesNo | showAlternative('Answer not provided') }}
                   </dd>
                 </div>
                 <div
@@ -778,7 +778,7 @@
                   <dd
                     class="govuk-summary-list__value"
                   >
-                    {{ application.canSpeakWelsh | toYesNo | showAlternative('Answer not supplied') }}
+                    {{ application.canSpeakWelsh | toYesNo | showAlternative('Answer not provided') }}
                   </dd>
                 </div>
                 <div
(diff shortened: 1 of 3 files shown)


src/views/Exercises/Applications/Application.vue
```diff
@@ -328,16 +328,25 @@
               </h2>
 
               <dl
-                v-if="application.equalityAndDiversitySurvey"
+                v-if="application.equalityAndDiversitySurvey && Object.keys(application.equalityAndDiversitySurvey).length"
```
warrensearle, line 331: It would be nice to simplify / minimise the amount of Javascript code within the template.

I wonder if there's a _simpler_ way to check what you need or perhaps use `computed` variable (without ending up with loads) or a helper method.

src/views/Exercises/Applications/Application.vue
```diff
@@ -952,44 +1000,44 @@
...
+            <div
+              v-if="
+                (exercise.appliedSchedule=='schedule-2-3' && application.applyingUnderSchedule2Three)
+                  || (exercise.appliedSchedule=='schedule-2-d' && application.applyingUnderSchedule2d)"
+              class="govuk-summary-list__row"
+            >
+              <dt
+                class="govuk-summary-list__key"
+              >
+                Explain how you've gained experience in law.
```
warrensearle, line 1029: The UI for this has gone a bit skew-wiff 
Screenshot 2021-03-04 at 11 06 29
Plus see that Memberships shows up when it's empty: so either needs text to explain or to be hidden

warrensearle (changes requested):
Please see comments

HalcyonJAC:
spotted a typo - `experience` is spelt wrong in this code block:

```
<div
    v-if="application.expereience && application.expereience.length"
>
```

`\src\views\Exercises\Applications\Application.vue`

src/views/Exercises/Applications/Application.vue
warrensearle: @tomlovesgithub Memberships still shows up when there's no data for it. Perhaps create a specific bug ticket to address this if you prefer.

warrensearle (approved):
Code looks ok - there's still room to simplify the template however no show-stoppers. 👍 
(oh I feel mean 😄)

tomlovesgithub:
Thank you Warren, very merciful  - I will return to slay the `applications.vue` one day

Issue #1198, opened 2021-03-02
Ex 051 - Unable to download QT - critical analysis test results


KateMJAC:
**The issue**

As an admin working on exercise JAC00051 – Deputy District Judge (Magistrates’ Court), I am unable to download the responses from the JAC staff dry run of our QT critical analysis test that was held on 1/03/21. 

We also held a JAC staff dry run QT situational judgement test on the same day and we are able to download those results. 

**Describe the solution you would like**

I would like to be able to download the results from the QT critical analysis test – JAC staff dry run held on 1/3/21.

mbrookeswebdev:
To test - 

1. Log into Admin, go to an active exercise, create a Qualifying test, invite test candidates. 
2. Log into Apply as test candidates, complete the test. 
3. Log into Admin, go to the created Qualifying Test, click on Responses and click on Export responses button. ([Apply test script](https://docs.google.com/spreadsheets/d/1ui5XkroLH1UHUgahLuaVrFLQr6afIuAXo1r9cQZRgeE/edit#gid=1815186770), [Admin test script](https://docs.google.com/spreadsheets/d/1BAyEPjLIOQcVFfSCRTsBzDL-lKeQscuFkTAWLJelcvU/edit?usp=drive_web&ouid=104315484718704594540))
4. Make sure it is possible to download responses, check negative behaviour (i.e.skipping questions in a test etc) as well as positive.
5. Make sure all data is present.

warrensearle:
Have reviewed code and approved - with a suggestion for simplifying.

Have also tested against the prod data and it enables ex 51 data to be downloaded

HalcyonJAC:
Hi @mbrookeswebdev @warrensearle  

If this code change is all ok can it be merged as a priority please?

I ask as we are working on the following item at the moment, which will effect this export/download:

#1100 - Exports should be prepared server-side

Locally, I have already created a feature branch off of develop for #1100. Let me know once this pull request has been merged though and I will merge develop into my local feature branch for #1100 to help avoid any merge conflicts coming up when its time to review #1100 ;)

Thanks

KateMJAC:
I have tested this on staging and was able to download CA QT results that included null values. When compairing the download with the responses held on the platform thay corresponded. Null values were displayed in the download as ---, clearly indicating that no answer was provided. 

Everything now works as it should. Thank you Maria.

PR #1202, opened 2021-03-03
#1198 Fixed issue of a QT results  download unable to start if candidates skip questions and submit with no answers.

mbrookeswebdev:
Testing instructions are in the ticket description along with test scripts for reference.

### DIFF
--- a/src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
+++ b/src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
@@ -161,7 +161,9 @@ export default {
             if (response) {
               const responseSelection = response.selection;
               if (responseSelection) {
-                if (responseSelection.mostAppropriate !== undefined && responseSelection.leastAppropriate !== undefined) {
+                if ((responseSelection.mostAppropriate !== undefined && responseSelection.mostAppropriate !== null)
+                  && (responseSelection.leastAppropriate !== undefined && responseSelection.leastAppropriate !== null))
+                {
                   row.push(
                     question.options[responseSelection.mostAppropriate].answer,
                     question.options[responseSelection.leastAppropriate].answer,
@@ -202,7 +204,7 @@ export default {
             }
             if (responses) {
               responses.forEach((response) => {
-                row.push(response.text === null ? 'Question skipped' : response.text);
+                row.push(response.text === undefined || response.text === null ? 'Question skipped' : response.text);
               });
             }
           });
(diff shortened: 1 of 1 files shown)


src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
```diff
@@ -161,7 +161,9 @@ export default {
             if (response) {
               const responseSelection = response.selection;
               if (responseSelection) {
-                if (responseSelection.mostAppropriate !== undefined && responseSelection.leastAppropriate !== undefined) {
+                if ((responseSelection.mostAppropriate !== undefined && responseSelection.mostAppropriate !== null)
+                  && (responseSelection.leastAppropriate !== undefined && responseSelection.leastAppropriate !== null))
+                {
```
warrensearle, line 166: You could perhaps simplify this by using `== null` to cover both the `null` and `undefined` scenarios. So
```
if (
  responseSelection.mostAppropriate != null && 
  responseSelection.leastAppropriate != null
)
```

src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
```diff
@@ -202,7 +204,7 @@ export default {
             }
             if (responses) {
               responses.forEach((response) => {
-                row.push(response.text === null ? 'Question skipped' : response.text);
+                row.push(response.text === undefined || response.text === null ? 'Question skipped' : response.text);
```
warrensearle, line 207: Perhaps use `!= null` to check for `null` and `undefined`

src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
```diff
@@ -219,7 +221,7 @@ export default {
             }
             if (response) {
               const responseSelection = response.selection;
-              if (responseSelection !== undefined) {
+              if (responseSelection !== undefined && responseSelection !== null) {
```
warrensearle, line 224: Perhaps use `!= null` to check for `null` and `undefined`

warrensearle (approved):
This is working, so have approved. 👍 
Please see comments re perhaps simplifying the code

src/views/Exercises/Tasks/QualifyingTests/QualifyingTest/Responses.vue
mbrookeswebdev: I try not to use == anymore and go for ===, to avoid any conversions taking place and make it clear as to what we need to check.

warrensearle: Yes it's best to use `===` but there's a camp, which includes me, that thinks it's ok to use `== null` and `!= null`.
e.g. see here: http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html
It's fine either way

PR #1205, opened 2021-03-08
#1192 Hide the Transfer Handover button until the epic work is complete

mbrookeswebdev:
As discussed with @KateMJAC, I've hidden the button until the JO Handover epic is complete, to avoid confusing users.


warrensearle (approved):
Approved! 👍 
(alternatively we could show this only for Super Admin users - so that we can see & use whilst it's being worked on)

PR #1220, opened 2021-03-12
Implemented Agency report (which is generated server-side)

HalcyonJAC:


mbrookeswebdev (commented):
I did click the button Refresh and saw this error

screenshot

HalcyonJAC:
@mbrookeswebdev, That's because I forgot to deploy the digital-platform changes to the _develop_ site! 

Please try again.

mbrookeswebdev (approved):

warrensearle (changes requested):
This is working well in terms of:
- [x] Can refresh the report
- [x] Can export report data

However the report data also needs to be displayed in the tabs on this page.
Please see the `agencyReport` getter in store/applications.js for where this is/was being pulled together (which _may_ help).

Also, for bonus points, it would be good to display the timestamp on the report page somewhere - e.g. Diversity report:
Screenshot 2021-03-15 at 14 35 15

HalcyonJAC:
@warrensearle I'm not sure what data is missing here. When I test using exercise ID 7eUXHjgVdBmol9qZ6xVD data is displayed on every tab.

warrensearle:
Hey @HalcyonJAC have tried this again. Ideally we should be able to view the tabs content and export the data without needing to load the applications documents. Right now the `created` method loads all the applications documents - see https://github.com/jac-uk/admin/blob/153fd0135cbefecbe7f68e4609a9834ef38072bc/src/views/Exercises/Show/Reports/Agency.vue#L488-L492

HalcyonJAC:
@warrensearle I see your point. I have just pushed another change up so that the page no longer binds to the applications collection at all. Instead it now inspects the report that has been generated and uses that to render the display

warrensearle:
@HalcyonJAC great... almost there 😄 
I _think_ this means the `agencyReport` code can be removed from here:
https://github.com/jac-uk/admin/blob/d38464a273b92dda7972afe8cc9d846b67bd662e/src/store/applications.js#L69

HalcyonJAC:
@warrensearle It does indeed! I have done a solution-wide search for `agencyReport` and it is not referenced anywhere else in the application, so have removed the redundant code.

warrensearle (approved):
Awesome, thank you @HalcyonJAC 👍

PR #1221, opened 2021-03-15
Feature/616 add first last names

mbrookeswebdev:
Implemented changes:

When the candidate's application form is viewed, Personal details should display first/last names if exist or a full name if they do not exist. Candidate's details page should also display first and last name and save updated details correctly (Personal Details component in the jac-kit was updated to reflect the changes).

Exercise-Application-JAC-Digital-Platform


src/views/Exercises/Applications/Application.vue
```diff
@@ -180,7 +180,46 @@
               </h2>
 
               <dl class="govuk-summary-list">
-                <div class="govuk-summary-list__row">
+                <div
+                  v-if="application.personalDetails.firstName && application.personalDetails.lastName"
```
warrensearle, line 184: Could perhaps put `application.personalDetails.firstName && application.personalDetails.lastName` in a computed as it is repeated three times in the template

warrensearle (commented):
Tested on the preview URL and this is working well.
For bonus points I wonder if there's a way to allow the user editing `fullName` to be able to switch to editing `firstName` and `lastName` instead?

HalcyonJAC:
@mbrookeswebdev 

The changes look ok to me but I am not really sure what the fullName vs. firstName and lastName split is really about to be honest. I am wondering though if they effect the reports that have been written at all? I ask because most of the reports reference `application.personalDetails.fullName`. For example: (see below):

image

src/views/Exercises/Applications/Application.vue
mbrookeswebdev: Done.

mbrookeswebdev:
@HalcyonJAC We need to collect first/last names to enable the handover of data to the Judicial Office. Whilst we will be collecting this data, full names of candidates will still exist, so it should not affect your work.

warrensearle (approved):
This looks good to me. Tested locally 👍

HalcyonJAC:
> @HalcyonJAC We need to collect first/last names to enable the handover of data to the Judicial Office. Whilst we will be collecting this data, full names of candidates will still exist, so it should not affect your work.

Gotcha! Thanks. :)

HalcyonJAC (approved):

PR #1222, opened 2021-03-18
Feature/1206 enable new and old character info on candidate page

mbrookeswebdev:
To test - please check Candidate view (Applications > any candidate > full info > click on candidate's name) displays correct version of Character Info depending whether old or new form was used and margins look ok.

CharInfo


src/views/Candidates/CandidatesView.vue
```diff
@@ -103,9 +157,13 @@ export default {
     getUserId() {
       return this.$route.params.id || '';
     },
+    getApplicationVersion() {
```
warrensearle, line 160: I think we could use a `_versionNumber` in the `characterInformation` documents to simplify this. Rather than looking at the exercise (if we have selected an exercise).
The version number would be set via Apply so a small/simple change would be needed there.

warrensearle (changes requested):

src/views/Candidates/CandidatesView.vue
mbrookeswebdev: I've added _versionNumber to Candidate. It shouldn't be an issue, because I don't think anyone applied yet using the new form.

warrensearle (approved):
👍

Issue #616, opened 2020-06-18
Add character checks request/return dates to timeline


YaaL:
Replace the single "character checks" date with request and return dates, similar to Independent Assessments.
These dates will be displayed on the character checks task page, eventually also to automate sending the requests.
Should display "overdue" if there are outstanding requests or candidates have not filled in the form by return date.

YaaL:
@ClaireTroughtonJAC can you have a look at this and confirm we don't need anything else?

PR #1230, opened 2021-03-25
#616 Added code to save full name from Candidate view

mbrookeswebdev:


mbrookeswebdev:
Can't test this without deploying the updated PersonalDetails jac-kit component, but I am still having a dilemma with how to make jac-kit available for testing on staging.

warrensearle:
> Can't test this without deploying the updated PersonalDetails jac-kit component, but I am still having a dilemma with how to make jac-kit available for testing on staging without it being released into production first.

Hey @mbrookeswebdev we can lock our production code to a specific version of jac-kit. We can then upgrade to a newer version of the kit as and when we're ready, or not at all. Looking at `master` branch this will involve a hotfix to change package.json to the following (say):
```
"@jac-uk/jac-kit": "0.0.31",
```
instead of:
```
"@jac-uk/jac-kit": "latest",
```
This will mean our production site will always use version 0.0.31 of the jac-kit, for example.

warrensearle (approved):
👍

Issue #1242, opened 2021-04-01
All candidates in stages tab moved to withdrawn Exercise 014 and 010


joy-ade:
Moving candidates statues from shortlisted  to selected resulted in all the candidate that were in selected stage also appeared in withdrawn stages

To set statuses at different stages on the digital platform, this does not work properly. If I click on a candidate to change their status as withdrawn, the digital platform marks all selectable candidates as withdrawn. This is not reflected on the staging tabs, the stages are correct.

For example, if I am moving candidates from Passed Sift to Selection Days, these candidates will show as shortlisted to selected. If at this stage I select a candidate as withdrawn and set the new status as withdrawn, this will then move all of those candidates to the withdrawn tab. Although on the status it shows as selected, these candidates will also appear under 'withdrawn'. 

![image](https://user-images.githubusercontent.com/44227249/113321079-246d5380-930b-11eb-88b5-336cbac5b8c2.png)

![image](https://user-images.githubusercontent.com/44227249/113320956-03a4fe00-930b-11eb-821f-8ac77b2c0b61.png)

KateMJAC:
Tested by user testing group today with a positive result. Can move to review.

PR #1247, opened 2021-04-06
#1242 Fix issue with incorrect applicants moved to withdrawn status

mbrookeswebdev:


mbrookeswebdev:
To test:
Please try moving several candidates from stage Shortlisted to stage Selected (Invited To Selection Day) and then a few other candidates to Withdrawn. Only correct candidates should be moved to Withdrawn.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "admin",
-  "version": "0.59.0",
+  "version": "0.59.1",
```
tomlovesgithub, line 3: is this going to be released as a hotfix?

tomlovesgithub (approved):
Looks right, just wondering how it should be released (hotfix/otherwise)

package-lock.json
mbrookeswebdev: In the platform-issues form it says it was not urgent.

warrensearle (approved):
Tested using code on `develop` and again with this code and can see that it fixes the problem 👍

Issue #257, opened 2020-02-07
Admins have access to the candidate assessments and or CVs


ClaireTroughtonJAC:
Is your feature request related to a problem? Please describe.

Admin staff need to access candidate assessments and or CV to share them with panel members

Describe the solution you'd like

- They need a central store to see them - preferably on google drive or somewhere in the platform.  
- They need to be printable
- They need to have the candidate name and reference number display if possible.  

Describe alternatives you've considered
Automating the three elements that the panel needs to see into one electronic document that can be shared on Google Drive that the panel access to review applications and create a report about the candidates.

warrensearle:
On the Application Details page there are now direct links to the files the candidate uploaded. This is a basic/initial fix in order to meet today's deadline. We will need to revisit in order to include
- [ ] moving the files across to google drive
- [ ] virus checking

PR #1476, opened 2021-08-19
#257 Refine database security rules and write more tests

HalcyonJAC:
## What's included?
This work fixes the email verification issue with third party (specifically Microsoft) authenticated accounts to the admin platform. The security rules require for all users that an email address is verified, and there is no way of telling on firestore where the request originated from.

Microsoft accounts do not set the email verified flag to true. It is a requirements that these are set to true if logging in via the admin platform, as users from the JAC and other government departments logging into the platform should not have to verify their email address.

This pull request is one half of the work to fix this issue, by checking the authentication provider and if the email verified flag is false, requesting a cloud function (with proper access to update all google authentication service user details) to verify their email.

The other half of this is https://github.com/jac-uk/digital-platform/pull/633.

This should also fix the case where an existing user has logged in - it'll verify their email when the SPA is loaded.

## Who should test?

- [x] Product Owner
- [x] Developers
- [ ]  UTG

## How to test?

Login using a microsoft account. Once the digital platform work is merged as well, it should work. Unfortunately the only way to tell if it doesn't work as if you login using a MS account, and it gives you permission denied.

🟠 **Medium risk - this does change code that is shared with other areas**

## Additional context
Include screen grabs, video demo, notes etc.

---
PREVIEW:DEVELOP


warrensearle (approved):
✅ Code review. Looks good to me

mbrookeswebdev (approved):

tomlovesgithub (approved):

warrensearle:
@HalcyonJAC @JeremykJAC looks like the corresponding digital platform work was reverted. It now exists in https://github.com/jac-uk/digital-platform/pull/642.
Can you get both parts approved & merged?

Issue #1356, opened 2021-05-28
QT Timer Issues


joy-ade:
There were issues with the timer while candidate were sitting for the QT test. The was discrepancies between our server timer and local time. As a  result of this few candidates were log out of of the test. @warrensearle has done a lot investigation and he  has also identified the solution.

@warrensearle can you all more information to this ticket please

while testing the QT on 12/7/2021  the following scenario were tested on the timer.

Maria set her time to one hour behind. When she log into her profile to take the test she was allocated 1hr 40mins. she had advantage of 1hour while sitting for test.![QTsc2.png](https://images.zenhubusercontent.com/5f7b3a91edd7a0a890566450/882428dc-41fe-4e07-9d62-dec347dfe4d5)

Tom also set his time and date to previous day and he was allocated 14hours to take the test.![image (8).png](https://images.zenhubusercontent.com/5f7b3a91edd7a0a890566450/0a05e1e0-1e62-40e4-819f-96da5ea2061e)

PR #1479, opened 2021-08-20
#1356 Display client details of QT participant

warrensearle:
## What's included?
During a QT we collect information about the user's platform, browser and local time. Here we are displaying that information in Admin.

## Who should test?
✅ Product owner
✅ Developers

## How to test?
On a qualifying test response page notice there is a new tab called 'Client'.
The 'Client' tab **should not** be available before the test has been started (as we haven't been able to collect data from the user).
The 'Client' tab **should** be available during and after a test has been started.
When available the 'Client' tab should display key information about the user's local environment such as:
- Platform
- Browser
- Timezone
- Initial time offset. This is the difference in seconds between the user's local time and our server's time
- Latest time offset. A significant change here from the initial value indicates that the user changed their local clock (or device) during the test.

## Additional context
Here's an example of the Client tab
image

---
PREVIEW:DEVELOP

### DIFF
--- a/src/views/Exercise/Tasks/QualifyingTests/QualifyingTest/Response/View.vue
+++ b/src/views/Exercise/Tasks/QualifyingTests/QualifyingTest/Response/View.vue
@@ -351,7 +351,62 @@
             </table>
           </div>
         </div>
-        <!-- // END CONNECTION TAB -->
+        <div v-if="activeTab === 'client'">
+          <h2 class="govuk-heading-m">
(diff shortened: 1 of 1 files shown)


mbrookeswebdev (approved):
👍  Awesome :)

HalcyonJAC (approved):

KateMJAC:
Merging for testing via dry run on Tue 24th August.

PR #1537, opened 2021-12-22
- Modal wont cancel, PQE blank fields bug, char info shows 'no info...' rather than no

tomlovesgithub:
issues addressed
- Modal wont cancel 
- PQE blank fields bug 
- char info shows 'no info...' rather than no 
- Inputing custom othertask overwrites tasks 
- a few places relied on `obj.change` which was meant as a check for the existence of the value, currently doesnt accommodate false values 

PREVIEW:DEVELOP

### DIFF
--- a/src/components/Page/InformationReviewSectionRenderer.vue
+++ b/src/components/Page/InformationReviewSectionRenderer.vue
@@ -223,12 +223,6 @@
     </div>
 
     <span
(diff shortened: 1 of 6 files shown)


warrensearle (approved):
Looks good @tomlovesgithub 👍

PR #1542, opened 2022-01-11
1494 - Character issues paging not working

tomlovesgithub:
## What's included?
The pagination was incredibly buggy on the character reports page.
I was unable to tell exactly what this was intended to do as it wasn't functioning.
i believe the basis of the issue was the `orderBy(status)` which was causing results to be differently ordered upon a page change, null-ifying the change. 

## Who should test?
✅ Developers

## How to test?
- Find an exercise with more than 10 character issues
- go to exercise > reports > character issues
- test pagination (next + previous)

---
PREVIEW:DEVELOP


src/views/Exercise/Reports/CharacterIssues.vue
```diff
@@ -284,9 +284,7 @@ export default {
       }
       if (this.candidateStatus !== 'all') {
         firestoreRef = firestoreRef.where('status', '==', this.candidateStatus);
-      } else {
-        firestoreRef = firestoreRef.where('status', '!=', 'withdrewApplication').orderBy('status');
-      }
+      }
```
warrensearle, line 287: I think you should just drop the `orderBy` from here but keep the `where` as this excludes withdrawn applications from the report.

warrensearle (dismissed):
See comment about excluding withdrawn applications from the report.

HalcyonJAC (approved):
I tested using https://jac-apply-admin-production--pr1542-1494-paging-not-wor-xulbd33f.web.app/exercise/i1JUb4lIjGxAXcembR09/reports/character-issues# and the pagination seems to work fine.

src/views/Exercise/Reports/CharacterIssues.vue
tomlovesgithub: removing only `orderby('status')` yeilds an error, i did try adding an additional orderby, but i couldnt work out the original ordering.

 "FirebaseError: [code=invalid-argument]: Invalid query. You have a where filter with an inequality (<, <=, >, or >=) on field 'status' and so you must also use 'status' as your first orderBy(), but your first orderBy() is on field 'candidate.fullName' instead."

warrensearle:
@tomlovesgithub have made a few fixes to the paging

warrensearle (approved):
👍

PR #1544, opened 2022-01-12
#1545 - Revise Character issues version logic

tomlovesgithub:
## What's included?
Revision of `isVersion2` computed, which had an oversight causing the wrong version to show.
Detail field in v1 character from on application view now only shows if true, or in edit mode.

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?
- Observe Character form is displaying correct Qs and As
- Ensure 'details' only show when an character issue has been marked as true

PREVIEW:DEVELOP


warrensearle (approved):
Looks good to me 👍

Issue #1616, opened 2022-05-16
Character Report - Statuses 


JeremykJAC:
Save the statuses:

 ![Screenshot 2022-05-16 at 10.00.39.png](https://images.zenhubusercontent.com/603ccb08852e7606029cef76/fa8c6776-aab1-41f7-ad48-5067008f573f)

Add a filter to the top of the page to allow a user to filer by status.

PR #1618, opened 2022-05-16
#1616 character report statuses

HalcyonJAC:
**Author checklist**

- [x] Include primary ticket number in title - e.g. "#123 New styling for widget" - and any additional tickets in description
- [x] Fill in the details below and delete as appropriate
- [x] Be proactive in getting your work approved 💪

---
## What's included?
Bug - The Character Issues page allows the user to choose a status to assign to each application but their choice does not get saved to the database.

1. Fix the bug 
2. Add an recommendation status filter (so the user can only see rejected applications, for example)

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?
preview URL: https://jac-admin-develop--pr1618-feature-1616-charact-2gnd95ly.web.app

- Load the character issues page
- Set recommendation status for each applications
- Reload the page. The chosen statuses should be visible still (i.e. they were saved to the database)
- Check the recommendation status filter works

## Additional context
Demo:

https://user-images.githubusercontent.com/79906532/171231938-c1a0f532-182f-404f-b70d-8d8256e76d56.mp4

## Related permissions
Have permissions been considered for this functionality?
- No permission changes required

---
PREVIEW:DEVELOP


warrensearle (changes requested):
Looks good however there is a glitch with the filtering:

e.g. 
If we are looking at all issues with status of Proceed then we should only see candidates which have one or more issue with status of Proceed. Right now we are seeing all candidates whether or not they have matching issues.
A screen grab might help me explain...

Here we can see all character issues for my exercise:
image

Here I want to only see issues marked as Proceed. There are none so I would expect an empty list but instead I see a candidate listed with no issues:
image

So...when we are filtered by status please only show candidates with issue(s) matching that status

warrensearle (approved):
Looks good to me 👍

warrensearle (approved):
Re-approved 👍

andrew-isaac (approved):
👍

andrew-isaac (approved):
👍

tomlovesgithub (approved):
Great stuff, tested functionality locally too!

PR #1654, opened 2022-06-24
Amend appCheck logic

tomlovesgithub:
## What's included?
The initialisation of the app check had some redundancy issues when the token isn't present. 
Changing the chronology and setting the appCheck constant with `let` means it will work with or without

## Who should test?
✅ Developers

## How to test?
???
- Ensure app check works with or without recapcha token.

---
PREVIEW:DEVELOP


warrensearle (approved):
@tomlovesgithub thanks for fixing the way App Check is loaded.

Happy to approve as is however just wondered if we really do want to change the functions emulator port as part of this fix?

Issue #1666, opened 2022-07-04
Stages and statuses


warrensearle:
Improve the stages and statuses that applications can be assigned upon their journey from 'registration' to 'handover'.

See [this google doc](https://docs.google.com/document/d/1b0bQ6O6rhyL2fhmZWq5OaL0owBnlaYvY/) for more information

**To do**
We will approach this work as follows:

**1. Stages and statuses**
- [x] Update the existing stages and statuses screens to support the new stages and statuses
- [x] Centralise the logic
- [x] Add custom page size
- [x] Add search
- [x] Enable EMP to be updated without changing status
- [x] Include a star next to statuses that progress application to next stage
- [ ] Update reports which include stage & status - e.g. diversity report  

**2. Back-fill data / support current and new stages & statuses**
In order to facilitate user testing and trial usage we will support current & new stages and statuses.
When all stakeholders are happy with the new stages and statuses we will back-fill / transform old stages and statuses into the new equivalents. This will help with reporting & stats across exercises.

 - [x] Assess the scope/scale of back-filling data vs supporting old and new model
 - [x] Support current and new stages & statuses
 - [ ] Back-fill / migrate data

andrew-isaac:
@warrensearle  - What is the update with this work? Why is it blocked?

PR #1673, opened 2022-07-07
#1666 Stages and statuses

warrensearle:
## What's included?
Introduces our new set of Stages and Statuses for applications.

Main ticket: #1666 

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?
Create or use an existing exercise which has applications.
Begin processing applications and navigate to Stages.

- Check that applications can be assigned the statuses and can progress through the amended stages
- Check that page size can be amended
- Check that search works
- Check that filter works

🟠 Medium risk - this does change code that is shared with other areas

---
PREVIEW:DEVELOP


andrew-isaac (commented):
👍 works for me! I have sent it to UT Group

PR #2059, opened 2023-07-11
#2058 Allow blank scenario scores => did not take test

### LINKED ISSUE #2058: Allow blank scenario scores => did not take test
**Problem**

Currently when scenario scores are imported or entered on screen we expect all candidates to be given a score. This is not actually the case as some candidates might not have taken part in the test.

We also need to distinguish between candidates receiving 0 score versus candidates who did not sit the test.

**Solution**

When importing scenario scores, or entering on screen, we should allow rows with no scores.

If a candidate has no score then it will be interpreted as they did not take the test.

If a candidate has 0 score then it means they took the test but did very poorly.

warrensearle:
## What's included?

Allows data entry to be completed even if some applications do not have scores entered into the platform. These will be interpreted as 'did not take test'.

Includes a modal to check that the user is happy to proceed even if some applications are missing scores.

Enables user to proceed to the merit list and setting a pass mark.

Closes #2058 

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?
For an exercise that has run a scenario test and has scores to be entered on platform, navigate to Tasks > Scenario test:

- When entering data for a scenario test leave some applications without scores.
- Press the Continue button and check that a warning modal is displayed.
- Confirm you'd like to continue even though some applications don't have scores.
- Check that you successfully proceed to the merit list screen (where pass mark and overrides can be configured)

Data entry screen:
image

Expected warning modal:
image

Expected destination:
image

---
PREVIEW:DEVELOP


nickaddy:
@warrensearle A couple of questions: does it make a difference whether I test this on Develop or Staging? Is there an easy way of finding an exercise that has a scenario test?

tomlovesgithub (approved):

HalcyonJAC (approved):
👍

PR #2081, opened 2023-07-21
#2079 Eligibility Checkbox Groups Validate Error

### LINKED ISSUE #2079: Eligibility Checkbox Groups Validate Error
When the Eligibility form is submitted the checkbox groups show an error implying they're empty when they're not.
See screenshot:

Screenshot 2023-07-21 at 09 01 55

drieJAC:
## What's included?
Updated to use fix in jac-kit. Altered checkbox group variables to be empty arrays by default instead of null.

Closes #2079 

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?

- Create a new exercise
- Go to the Eligibility tab and fill in the form fields and save with no errors
- Click back into the Eligibility tab
- Click the Update Eligibility Information link
- Scroll to the bottom and press the Save And Continue button
- Ensure there are no errors for the Authorisation field
- Deselect everything in the Authorisation field and press Save And Continue and ensure an error displays telling the user they need to fill in this field

---
PREVIEW:DEVELOP


warrensearle (commented):
It would be very preferable to find a fix that's self-contained within the `CheckboxGroup` component. If we can't then we'll need to update the defaults used for all instances of the `CheckboxGroup` which is around 15 or so uses.
That update should be part of this fix.

drieJAC:
> It would be very preferable to find a fix that's self-contained within the `CheckboxGroup` component. If we can't then we'll need to update the defaults used for all instances of the `CheckboxGroup` which is around 15 or so uses. That update should be part of this fix.

Yeah I removed the bit of code that set the initial values to empty arrays instead of null and it works ok.

warrensearle (approved):
👍

HalcyonJAC (approved):

NormaJAC:
Error message displayed as expected

PR #2256, opened 2023-12-15
Feature/2252 User Management Comms & Search functionality

### LINKED ISSUE #2252: User Management Comms & Search functionality
**User Story**
As the Digital team, I would like new users to be emailed with confirmation of account set up and login details so that they are able to log in and start using the platform.

**Feature**
- Create a Notify template containing links and guidance: Done - [link](https://www.notifications.service.gov.uk/services/0abe6c8e-0b87-4cde-9493-5da4921ccc53/templates/4f221946-629e-468d-a83e-dec3d4d3d7d2)
- Write Cloud function that is triggered when the new user invitation is created
- Add surname as a searchable field for the search box on the user management page
- Order the users by surname

**Comments**
I have create the Notify template, but not set up the API. It currently has only 2 custom fields ((email)) and ((username)); not sure if the nomenclature corresponds with the relevant field set up in the database. I'm thinking that maybe it might be easier to just trigger a single email when the invitation is created on Production and the email includes info about all 3 environments, otherwise we'd have to set up a link and details database fields for each environment (unless one already exists?)

HalcyonJAC:
## What's included?
Closes #2252 

- [x] Add display name and email as a searchable field for the search box on the user management page.
- [x] Order the users by their last name.
- [x] Fix the wording of the error message.
- [x] Check if `getIdTokenResult` exists before invoking it (It might be empty during the first sign-in using a Microsoft email address). 

Note: This PR needs changes in [digital-platform: #2252 User Management Comms & Search functionality](https://github.com/jac-uk/digital-platform/pull/987).

## Who should test?

✅ Product owner
✅ Developers
✅ UTG

#### Sorting and searching in the user list
1. Go to "Users".
2. Check if the users are ordered by their last name.
3. Check if "display name" and "email" are searchable.

#### User invitation notification
1. Go to "Invitations" under "Users" section.
2. Click "Create" button to create a user invitation.
3. Go to "Notifications" and check if the notification is created.

## Additional context
Include screen grabs, video demo, notes etc.

## Related permissions
Have permissions been considered for this functionality?
- No permission changes required
- Permissions have been added / updated. Details:

---
PREVIEW:DEVELOP


warrensearle (changes requested):
@HalcyonJAC see comments on the digital platform PR. The proposed PR parses the display name, infers first name and last name and then stores them in our database. This feels a bit fragile and likely to be problematic in the future.
We can support search and sort using display name as is.
If we need more then we should use Microsoft API directly

nickaddy:
@HalcyonJAC @warrensearle Can we have a quick catch-up tomorrow to confirm which accounts are being/should be filtered here? I'd like to address these questions:

1. Presumably we only want to include here the accounts that have permissions for the 3 Admin environments?
2. Do the .digital accounts listed actually relate to Apply rather than Admin because it's my understanding that only Digital Super Users are able to access the Admin platform with .digital addresses. 
3. There's a whole of users here (with .digital addresses) that are listed as Super Users, but actually should be Ops team members or other; what's going on with that?
4. The sorting doesn't quite work, I believe due to the way accounts are set up, but don't think this will be a problem if we are only viewing .gov.uk accounts. Besides, we have the search function.
5. In terms of changing a user permission, is the change made when I click off the dropdown as there is no submit button?
6. What else do we need to do to get this live and test on production?

nickaddy (changes requested):
Filtering and ordering - great.
Surname search - not working in develop (only needs to work for .gov.uk accounts)

drieJAC (approved):

warrensearle (approved):
👍

nickaddy (approved):
Looks great, thx Ryan

PR #2264, opened 2023-12-19
Fixed an error in candidate submission due to there being no screensh…

drieJAC:
Fixed an error which was caused when submitting a an issue from a candidate.
The feedback form should now be error free.

---
PREVIEW:DEVELOP


Franceswog (approved):
Works great 
image

PR #2415, opened 2024-06-03
Statuses for withdrawn candidates

tomlovesgithub:
## What's included?
Updates to `store/application.js` so that a check for an accompanying `applicationRecord` can be accessed and used when withdrawing an application from the application view. 
Using this, the button now updates the withdrawn status to both if both exist, if not, only the application status is updated. 

## How to test?
Withdraw an application from an application view. Ensure that the status is updated and reflected in all application lists.

## Additional context
image

---
PREVIEW:DEVELOP

### DIFF
--- a/src/store/application.js
+++ b/src/store/application.js
@@ -1,8 +1,9 @@
-import { collection, doc, updateDoc, setDoc, addDoc, serverTimestamp, runTransaction } from '@firebase/firestore';
+import { collection, doc, updateDoc, setDoc, getDoc, addDoc, serverTimestamp, runTransaction } from '@firebase/firestore';
 import { firestore } from '@/firebase';
 import { firestoreAction } from '@/helpers/vuexfireJAC';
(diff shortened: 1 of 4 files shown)


nickaddy (changes requested):
@tomlovesgithub The results seem to be patchy - see the candidates under Withdrew on [this exercise](https://admin-develop.judicialappointments.digital/exercise/GYysuH7M4V21MFVM8Dzg/applications/withdrawn). Some have their status correctly populated, others don't. I've tried to isolate whether the unpopulated cases are either from pressing the button or changing the status but there doesn't seem to be any rhyme or reason.

nickaddy:
Looking good, Mr T! I note there are still a few candidates on 184 with a blank status column. Once you've addressed, I'll be happy to approve the ticket :-)

nickaddy (approved):

drieJAC (approved):

HalcyonJAC (approved):

PR #2454, opened 2024-06-21
Make Pagination numerical by default

tomlovesgithub:
## What's included?
do-over of old branch with merge issues.

Alphabetical pagination on the stat-con report page was causing confusion with the stat-con download. 
I have made the pagination numerical by default so this page should always load any results, reducing confusion. 

## Who should test?
✅ Product owner
✅ Developers
✅ UTG

## How to test?
visit stat con page (exercise >> reports >> statutory-consultation)
ensure pagination is numerical on initial load.

---
PREVIEW:DEVELOP

### DIFF
--- a/src/views/Exercise/Reports/StatutoryConsultation.vue
+++ b/src/views/Exercise/Reports/StatutoryConsultation.vue
@@ -158,7 +158,7 @@ export default {
         where('status', 'in', statuses)
       );
       params.orderBy = 'candidate.fullName';
-      const res = await tableAsyncQuery(this.applicationRecords, firestoreRef, params);
+      const res = await tableAsyncQuery(this.applicationRecords, firestoreRef, params, null);
       firestoreRef = res.queryRef;
       this.total = res.total;
       if (firestoreRef) {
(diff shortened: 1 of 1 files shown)


src/views/Exercise/Reports/StatutoryConsultation.vue
```diff
@@ -114,6 +115,7 @@ export default {
       unsubscribe: null,
       unsubscribeReport: null,
       report: null,
+      total: null,
```
drieJAC, line 118: maybe best to set this to 0 instead of null

src/views/Exercise/Reports/StatutoryConsultation.vue
```diff
@@ -156,7 +158,9 @@ export default {
         where('status', 'in', statuses)
       );
       params.orderBy = 'candidate.fullName';
-      firestoreRef = await tableQuery(this.applicationRecords, firestoreRef, params);
+      const res = await tableAsyncQuery(this.applicationRecords, firestoreRef, params);
```
HalcyonJAC, line 161: Could you pass `null` as the 4th parameter, please? It will be like:
`const res = await tableAsyncQuery(this.applicationRecords, firestoreRef, params, null);`

HalcyonJAC (changes requested):

drieJAC (approved):

PR #2678, opened 2025-01-20
Bugfix: Permissions for 'Create demo applications'

warrensearle:
## What's included?
Recent changes to tighten user permissions on cloud functions and UI features had introduced, or highlighted, a bug with the 'Create demo applications' buttons. There are two buttons with this label and only one had been updated with the correct permissions. This small PR fixes that.

The 'Create demo applications' button should only be displayed for user who have the `canCreateTestApplications` permission in their user role:

image

## Who should test?
✅ Product owner
✅ Developers

## How to test?
You'll need a published approved exercise which does not have any applications. This one might work (for a while until someone creates applications!): 
https://jac-admin-develop--pr2678-bugfix-permissions-c-gxp3i72z.web.app/exercise/irlE2FKZJnFLOYAIsRHx/dashboard

Check the following:
 - If the logged in user does not have permission to create applications, you cannot see the 'Create demo applications' button
 - If the logged in user does have permission to create applications, you can see the 'Create demo applications' button

Compare this with the same checks on admin develop (i.e. without this fix) where the button displays if the user doesn't have permission and then disappears:
https://admin-develop.judicialappointments.digital/exercise/irlE2FKZJnFLOYAIsRHx/dashboard

---
PREVIEW:DEVELOP

### DIFF
--- a/src/views/Exercise.vue
+++ b/src/views/Exercise.vue
@@ -57,7 +57,7 @@
             class="govuk-!-margin-top-1"
           >
             <button
-              v-if="isReadyForTesting && !isArchived"
+              v-if="isReadyForTesting && !isArchived && hasPermissions([PERMISSIONS.applications.permissions.canCreateTestApplications.value])"
               class="govuk-button govuk-button--secondary"
               @click="changeNoOfTestApplications()"
             >


nickaddy (approved):
Cheers, Warren

tomlovesgithub (approved):

